PR #2, opened 2017-10-27
types: fix encoding of large numeric node ids

florolf:
Currently, NodeIds of the "Numeric" type are encoded as a UInt16
NamespaceIndex and a UInt64 identifier. However, according to the spec
(Table 6 of Part 6), a UInt32 should be used for the identifier.

Fix the en-/decoder accordingly and add a test.

From my understanding of the standard, one could also argue to reduce the integer in the `Identifier` enum to an `UInt32`, but I'm not sure if this is actually correct.


locka99:
Thanks for the patch. I'm currently away from my development PC so I can't commit this yet. But I'll take a look early next week.

PR #25, opened 2019-02-16
use source environment settings to set exports

biancode:


locka99:
I modified it slightly because I had updated the file a little bit more yesterday

PR #26, opened 2019-02-26
Request the maximum possible session timeout when creating a session.

ColinFinck:
The OPC UA server always revises the session timeout, but previously adjusted it to its configured minimum timeout.
Now the timeout is adjusted to the server's maximum session timeout.
As long as no logic is in place that takes revised_session_timeout into consideration and sends periodic keepalives, this is a simple fix to combat many cases of silently dropped sessions.

### DIFF
--- a/client/src/session.rs
+++ b/client/src/session.rs
@@ -528,7 +528,7 @@ impl Session {
             session_name,
             client_nonce,
             client_certificate,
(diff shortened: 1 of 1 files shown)


locka99:
I'll take this for time being but it should probably be a configurable setting with a sensible default

PR #36, opened 2019-09-24
remove useless Send and Sync implementations.

nkbai:
remove useless Send and Sync implementations.
Send and Sync is automatically implemented when the compiler determines
it's appropriate.


locka99:
OpenSSL crates must have made their types implement Send so this code is no longer necessary

nkbai:
yes , foreign_type_and_impl_send_sync this macro implements Send and Sync.

PR #49, opened 2020-05-06
types: Add missing AttributeId values

laumann:
OPC-UA spec Part 6, table A.1 gives a list of identifiers assigned to
attributes, where five attribute IDs were missing.

These were discovered with UaExpert that appears to explicitly request
some of these.

### DIFF
--- a/server/src/services/attribute.rs
+++ b/server/src/services/attribute.rs
@@ -425,6 +425,11 @@ impl AttributeService {
                 AttributeId::Historizing => write_mask.contains(WriteMask::HISTORIZING),
                 AttributeId::Executable => write_mask.contains(WriteMask::EXECUTABLE),
                 AttributeId::UserExecutable => write_mask.contains(WriteMask::USER_EXECUTABLE),
+                AttributeId::DataTypeDefinition => write_mask.contains(WriteMask::DATA_TYPE_DEFINITION),
+                AttributeId::RolePermissions => write_mask.contains(WriteMask::ROLE_PERMISSIONS),
+                AttributeId::AccessRestrictions => write_mask.contains(WriteMask::ACCESS_RESTRICTIONS),
+                AttributeId::AccessLevelEx => write_mask.contains(WriteMask::ACCESS_LEVEL_EX),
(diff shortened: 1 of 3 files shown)


laumann:
Hi @locka99! I'm not sure this is the right approach to get these values defined (or if it's even necessary). UaExpert functions fine without these attributes being available, so another "workaround" for this could also be to change the warning:

https://github.com/locka99/opcua/blob/147b98227c991481da38187a13502b37d38cb643/server/src/services/attribute.rs#L366

to a `debug!()`.

A little back-story: Using UaExpert, I was seeing warnings like these:
```
WARN: /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.7.0/src/services/attribute.rs:125: Attribute id 24 is invalid    
WARN: /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.7.0/src/services/attribute.rs:125: Attribute id 25 is invalid 
WARN: /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.7.0/src/services/attribute.rs:125: Attribute id 26 is invalid 
WARN: /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.7.0/src/services/attribute.rs:125: Attribute id 27 is invalid
```

and then tracked down this table that defines them:

image

locka99:
Thanks for the pull request. I think these attributes might have come in with OPC UA 1.04 and may be optional or be required to return valid values if they exist. Let me take a day to have a chance to review the change.

laumann:
Thanks, my guess is that they're optional, they do not cause any problems with UaExpert. And I think you're right, something valid probably needs to be returned.

If it's strictly 1.04, is there a way to indicate that when building? Or is that not necessary?

locka99:
Yes this seems to be a 1.04 thing. I'll merge this because all the nodes should return None for unsupported attributes and it fixes the attribute service. OPC UA Part 3 Section 5.9 has a table showing these to be optional args.

laumann:
Nice! Thanks!

PR #64, opened 2020-10-02
GUI for real-time data visualization

tushar-kamra:
For more information, please have a look at README.md

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -1,0 +1,16 @@
+## Demo GUI for real-time data visualization
+
+>### Requirements:
+1) Simulation server (Eg. Prosys simulation server)
+2) IDE with python libraries installed
+3) MySql database
+4) OPCUA Python library
+
+>### Features:
+1) This script enables you to visualize the OPC values in real-time with the facility of data logging. 
+2) An alert will be generated when the downtime reaches a threshold limit.
+3) Values can be adjusted as per the need.
+
+>### Screenshot of GUI:
+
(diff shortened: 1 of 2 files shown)


locka99:
Hi, I have a directory 3rd-party/ for 3rd party implementations of OPC UA. If you can produce a sample that is usable against one of the Rust samples I can include it in there. e.g. if your client can subscribe to a variable in demo-server or simple-server then it could be added. 

Primarily I'm interested in 3rd party OPC UA that allow me to split problems in half, e.g. if I have a bug in opcua-client then I'd like to be able to test simple-client against a 3rd party server. Conversely if I have a bug in opcua-server then I'd like to be able to test simple-server or demo-server against a 3rd party client. In that way I can halve where the bug may lie, in my code or some other implementation.

tushar-kamra:
A client can easily add more variables based on their usage. We just have to replicate the changes. A Client can scale this up to their needs.

locka99:
Hi, I'm going to close this because the functionality should exist in its own repository. If you need to reference this project then you can through a cargo file. I'm only taking stuff into 3rd party that allows me to swap out a simple-client or simple-server to test against the implementation. I would be open to a simple graphical user interface (e.g. like OPC UA Commander in the NodeJS OPC, or other clients like UA Expert) that utilised the Rust implementation as its backend, but it would have to be written in Rust and with some kind of cross-platform GUI toolkit so it could run on Linux and Windows

PR #66, opened 2020-10-08
Make it possible to use custom named .pem certs

svanharmelen:
We have a setup where we need to use specific client certificates which are fetched from a remote API during startup. These certs are in`PEM` format, so that wasn't possible with the current code. I tried to keep things as inline with the existing code as possible and made sure that the defaults would not change anything.

Additionally I made a small tweak to how certs are written to disk. The (3rd party) OPC server we connect to returns a certificate path which looks like `%BRAND%/UA Server [%UUID%].der`. So in order for this cert to be able to be saved, the directories `trusted/%BRAND%` need to exist. So now when writing a certificate to disk, the path needed is always being created just before writing the cert.

I hope this PR fits your style and vision and that this can be merged. Please let me know if you have any concerns or comments that need to be resolved first.

Thanks!

### DIFF
--- a/client/src/builder.rs
+++ b/client/src/builder.rs
@@ -114,6 +114,22 @@ impl ClientBuilder {
         self
     }
 
+    /// Sets a custom client certificate path. The path is required to be provided as a partial
+    /// path relative to the PKI directory. If set, this path will be used to read the client
+    /// certificate from disk. The certificate can be in either the .der or .pem format.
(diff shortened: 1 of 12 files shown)


svanharmelen:
@locka99 is this something that could be merged? I've been using it for a few weeks now and it works perfectly. So curious what would be needed to get this in...

locka99:
Yes, the code looks okay so I'll take it in there. Thanks!

svanharmelen:
Thanks @locka99!

PR #69, opened 2020-10-21
remove shutdown because it's useless

nkbai:
remove shutdown because it's useless.
Connection will be closed after return of spawn_writing_loop_task.
if add await here, it will hold a MutexGuard across await, and it's a compiler error.


locka99:
I added shutdown tests in several places because I couldn't force the server to abort listening for connections and I needed that to implement the behaviour in running tasks. That was tokio 0.1 of course, but are you saying the behaviour now can be done some other way? 

The server has an optional web interface that allows the server to be aborted with a call to /server/abort so it would be worth testing that functionality still works.

nkbai:
```
    /// Splits a `TcpStream` into a read half and a write half, which can be used
    /// to read and write the stream concurrently.
    ///
    /// Unlike [`split`], the owned halves can be moved to separate tasks, however
    /// this comes at the cost of a heap allocation.
    ///
    /// **Note::** Dropping the write half will shutdown the write half of the TCP
    /// stream. This is equivalent to calling `shutdown(Write)` on the `TcpStream`.
    ///
    /// [`split`]: TcpStream::split()
    pub fn into_split(self) -> (OwnedReadHalf, OwnedWriteHalf) {
        split_owned(self)
    }
```

these codes are part of tokio's stream.rs.

**Note::** Dropping the write half will shutdown the write half of the TCP stream.

nkbai:
```
impl Drop for OwnedWriteHalf {
    fn drop(&mut self) {
        if self.shutdown_on_drop {
            let _ = self.inner.shutdown(Shutdown::Write);
        }
    }
}

```

PR #70, opened 2020-10-25
Tokio 0.2

nkbai:
1. tokio_compat is back. rt.block_on not equals tokio_compat::run
2. add WaitGroup, but it's a little hard to use for complex tasks.
3. samples/modbus-server  upgrade to tokio 0.2
4. fix a bug for samples/modbus-server, nodeId conflict.

but it seems there is some error on subscription task of server.


nkbai:
I think the server API can be fully designed to be asynchronous, and the client side can provide both asynchronous and synchronous Apis.

Is this in line with your design goals?

locka99 (approved):

locka99:
Thanks for the changes I'll keep periodically merging these changes into my tokio-0.2.

PR #73, opened 2020-11-12
Try to properly handle session errors

svanharmelen:
@locka99 when testing our application we noticed that sometimes our session got closed by the (proprietary) server that we use, but that the client didn't handle the fact that the session was now actually closed. This resulted in a lot of errors like:

```
[DEBUG opcua_client::session] session:2 Service fault received with BadSessionClosed error code
...
[DEBUG opcua_client::session] session:2 Service fault received with BadSessionIdInvalid error code
```

But other then these messages, nothing happened to fix the session so the error state remained until the end of times :wink: 

When looking into the code we think we found the location were these kinds of errors should be handled, but are not sure what the best solution would be. In this PR I only called the session close callback and in our own code we now disconnect the session and try to connect again to get a new fresh session.

In the logs we can see that until our callback is executed the client keeps trying to send publish requests and keeps receiving the same errors, so it would probably be better if we can close, cleanup or disconnected the session as soon as we encountered these errors?

What do you think would be the best approach to solve this? Is calling the callback and trusting that the caller handles the problem the best solution, or is there something we can all in the client package itself that would be a better fit for the situation/problem? Thanks!

### DIFF
--- a/client/src/session.rs
+++ b/client/src/session.rs
@@ -2543,10 +2543,18 @@ impl Session {
                 let service_result = response.response_header.service_result;
                 session_debug!(self, "Service fault received with {} error code", service_result);
                 session_trace!(self, "ServiceFault {:?}", response);
-                // Terminate timer if
-                if service_result == StatusCode::BadTooManyPublishRequests {
-                    // Turn off publish requests until server says otherwise
-                    wait_for_publish_response = true;
+
+                match service_result {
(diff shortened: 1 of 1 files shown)


svanharmelen:
@locka99 really curious if this seems like a good fix, or that there are better solutions for catching and handling these kind of errors... So would be great if you have a minute to give some feedback on this one. Thanks!

svanharmelen:
Thanks!

Issue #74, opened 2020-11-18
cargo fmt --all?


svanharmelen:
I noticed that there is a `rustfmt.toml` file in the root, but if I try to format all code it gives me errors on that file:

<img width="697" alt="image" src="https://user-images.githubusercontent.com/4171547/99602063-10325500-2a01-11eb-81ea-75818e2a9df0.png">

I also noticed that the tokio-0.2 branch now contains a different `rustfmt.toml` file which does work, but causes all the code to formatted differently than on master.

This makes it hard to rebase the tokio-0.2 branch and to see what actual changes were made to upgrade Tokio. So would you be open to a PR that reformats and rebases both branches without using a custom `rustfmt.toml` (so just following the defaults)?

Of, if not, would you be open to add the same `rustfmt.toml` file used in the tokio-0.2 branch to master (including a reformat based on that)?

Last but not least, if you do have a strong preference, could you maybe supply a `rustfmt.toml` that can be used as a standard for this project?

Either one of these 3 solutions would make contributing, reading the code and understanding certain PR's a lot easier. I'm fine with doing the work needed btw... Thanks!

svanharmelen:
Guess this one can be closed as `master` is now formatted 🎉

PR #75, opened 2020-11-19
Reformat all the code using cargo fmt —all

svanharmelen:
This is a potential solution for issue #74, so trying to make all the code formatting consistent. Really hope this format (the default) works for you, but if not then please let me know what you would like to have changed. Thanks!


locka99:
Can you send me a merge with just the rustfmt changed? I'll incorporate that, although traditionally I've used the built-in formatter in IntelliJ so I need to see what impact it has on auto-generated code before I apply it to the rest of the source.

svanharmelen:
If you really want I will remove the second commit... But to be honest it doesn't do anything else then formatting  and cleaning up the imports. There were some strange (unneeded) imports like `use tokio::{self}` or `use std::{self, ...}`. And there were a couple of places were primitives like `u32` or `u64` were specifically imported, while in other places they were not (which makes sense as they are already available through the `std::prelude`).

So sure you want me to delete that second commit?

As for Intelij, I use that as well (CLion). In one of the latest plugin updates they added this one:

image

So that makes it easy to follow the community styling 😉 

Thanks for the quick reaction by the way! These types of PR's are usually nasty to keep in sync, so thanks for not letting that happen!

locka99:
u32 and u64 were imported for some min/max stuff which at the time required the import and maybe don't now. What I'll do today is bring the rustfmt in from your req and run it locally and hopefully that will reduce this request to zero

locka99:
I have done the following in master and will close this request:

1. Taken the rustfmt.toml from here (blank)
2. Changed all the autogenerated files to say #![rustfmt:skip] at the top because changing the scripts to output rustfmt friendly code is a pain
3. Run "cargo fmt" on the handwritten files

Concerning rustfmt I'm not sure I like the default formatting rules for a few things, especially the way an else doesn't start on a new line but appears as "} else {" for the time being I'll leave it this way.

svanharmelen:
Super nice! Thanks @locka99 👍🏻 

As for the defaults, you can of course always tweak the `rustfmt.toml` config to your preferences. Since the `rustfmt.toml` file is now in this repo, it will be used automatically by most (if not all) editors out there. So it will be much easier to keep the formatting consistent. Thanks again for doing this!

PR #76, opened 2020-11-19
Rebase on the updated (reformated) code in master

svanharmelen:
This PR cannot be merged, but has to be force pushed. It's a rebase on the reformatted code offered in PR #75 (which should be merged first). As there were a lot of merge conflicts it's not a clean update. Yet I confirmed that all changes are added correctly and that all tests still pass.

Again, hope that this format works for you but if not then please let me know how you want it. Thanks!


locka99:
See merge request #75. At present tokio-0.2 is receiving changes from @nkbai so for the time being I don't want to disrupt it.

svanharmelen:
Ah... I see...

@nkbai are you around? Are you willing to push these changes in your branch so this `tokio-0.2` branch is (more or less) in sync again with `master`? If you want I can make a PR to your forked branch instead is that is easier?

As for the rest @locka99, I'm in desperate need of a full async client that does not spin it's own runtime(s) (or threads for that matter) but uses an existing runtime.

We need this for an application we've build that uses your client, but we are encountering issues because we need to spin up quite a few separate/independent clients which now all start their own runtimes and threads which causes the application to run over 1024 file handles (as Tokio creates unix sockets for communication between the threads it manages).

So today I was doing pre-work for tomorrow and early next week, were I'm going to try to write such a version of the client. My idea was to make a copy of the existing client on this branch and then call it `async-client` that way both clients can evolve next to each other, but at the same time all the code can stay together in your repo.

But now reading your response I doubt if that is a good plan and if maybe I should just make a dedicated (separate) crate (still based on this client of course) instead (as this branch is a dedicated WIP branch of @nkbai)? Any thoughts or suggestions @locka99?

nkbai:
Okay, I’ll do a sync this weekend, and when I do, I want you to see if any other errors have been introduced.


There have been a lot of things going on lately, so I haven’t redesigned the async-client.

svanharmelen:
Hi @nkbai! Sounds good to me... Just let me know what and were to check 👍🏻 

I am actually already using the tokio-0.2 branch and found at least one issue in the updated code. But I will check things again after you rebased/synced. I would almost think that just copying the code from this PR (or force-pushing it onto your branch) is the easiest way as the rebase itself (which I done for this PR) caused a lot of merge conflicts I had to manually fix. But that is up to you of course.

And about the client... How do you feel about me adding a new `async-client` crate next to the existing `client` crate? We could maybe use this same branch, but that way they can exist side-by-side while developing them. As mentioned I am going to write one today and next week to at least solve our own problem. Of course it would be nice if I can contribute that client back it we can keep it in this public repo.

locka99:
Closing this also

PR #78, opened 2020-11-23
Make sure to stop when the sender end is dropped

svanharmelen:
When a channel does not have any senders, you will get a specific error indicating that the channel got disconnected.


locka99:
Thanks for the update!

PR #81, opened 2020-12-07
Use the current thread runtime for Tokio tasks

svanharmelen:
The default runtime is a threaded runtime which spawns a background
thread running a reactor instance and then creates a threadpool (with
the number of threads equal the the number of available CPU cores).

Now this creates a lot of threads, especially when multiple runtimes
are started, like in the client package. And while having a lot of
threads should not be a problem, Tokio needs to be able to communicate
between these threads so it (on Linux) creates unix pipes for that.
On an 16 core machine this resulted in around 144 handles per client.

On our system and the way we use multiple clients in our application
this caused us to hit the 1024 default open file handle limit. In most
cases that limit can easily be updated, but in our case that isn't an
option/solution. So I went looking at the code to see if there was an
"easy" fix and it seems there is one... As all the runtimes that are
created are actually created inside new dedicated threads, it seems
that using that single thread for the new runtime is a more efficient
solution which seems to be inline with how things were intended to
work.

I made the required changes in this PR and hope this is indeed inline
with how you intended to use Tokio. I excluded the runtime in the main
`run_server` function as I can imagine in that particular case it could
make sense to use a threaded runtime (now sure, can of cause make the
change for this one as well if you want).

Super curious what you think about this and if this is an acceptable PR.

### DIFF
--- a/server/src/http/mod.rs
+++ b/server/src/http/mod.rs
@@ -97,6 +97,7 @@ pub fn run_http_server(
     server_state: Arc<RwLock<ServerState>>,
     connections: Arc<RwLock<Connections>>,
     server_metrics: Arc<RwLock<ServerMetrics>>,
+    single_threaded_executor: bool,
 ) {
     let address = String::from(address);
     let base_path = PathBuf::from(content_path);
@@ -146,9 +147,16 @@ pub fn run_http_server(
 
     // Spawn a tokio task to monitor for quit and to shutdown the http server
     thread::spawn(move || {
-        tokio::run(quit_task.map(move |_| {
-            info!("HTTP server will be stopped");
-            let _ = addr.send(server::StopServer { graceful: false });
-        }));
+        let task = {
+            quit_task.map(move |_| {
+                info!("HTTP server will be stopped");
+                let _ = addr.send(server::StopServer { graceful: false });
+            })
+        };
+        if !single_threaded_executor {
+            tokio::runtime::run(task);
+        } else {
+            tokio::runtime::current_thread::run(task);
+        }
     });
 }
--- a/client/src/builder.rs
+++ b/client/src/builder.rs
@@ -236,6 +236,13 @@ impl ClientBuilder {
         self.config.session_timeout = session_timeout;
         self
     }
+
+    /// Configures the client to use a single-threaded executor. The default executor uses a
+    /// thread pool with a worker thread for each CPU core available on the system.
+    pub fn single_threaded_executor(mut self) -> Self {
+        self.config.single_threaded_executor = true;
+        self
+    }
 }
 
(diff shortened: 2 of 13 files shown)


locka99:
I think we should put this into a helper function and control the behaviour from configuration so the runtime can be fixed to 1 thread or many.

svanharmelen:
Sounds good... If you can point me to an example in this repo so I can try to use the same pattern? If not I'll try to come up with something myself 👍🏻

locka99:
The server has a config.rs and a builder.rs containing a ServerConfig and ServerBuilder respectively. Likewise for the Client and ClientConfig and ClientBuilder.  So if we added a single_threaded_executor property (defaulting to false) to each side then someone could set it like so:

```
Server server = ServerBuilder::new()
    .single_threaded_executor()
    .server()
    .unwrap();
```

Similarly for the client. 

Then in the places that create the executor could be passed the flag from somewhere that has the config to decide which executor to use.

svanharmelen:
Thanks, I'll have a look at this one later today!

svanharmelen:
@locka99 I just found some time to update this PR according to your feedback. Please let me know if this works for you, or that you would like additional changes. Thanks!

svanharmelen:
Rebased this one to fix the merge conflict introduced by merging PR #87

server/src/http/mod.rs
```diff
@@ -146,9 +147,16 @@ pub fn run_http_server(
 
     // Spawn a tokio task to monitor for quit and to shutdown the http server
     thread::spawn(move || {
-        tokio::run(quit_task.map(move |_| {
-            info!("HTTP server will be stopped");
-            let _ = addr.send(server::StopServer { graceful: false });
-        }));
+        if !single_threaded_executor {
+            tokio::runtime::run(quit_task.map(move |_| {
```
locka99, line 151: Can you assign this closure to a variable, e.g. stop_server_task because the if/else are running the same closure.

svanharmelen: Yeah, agree that would be better 👍🏻

svanharmelen:
Updated to make it a little more DRY

svanharmelen:
Hi @locka99! Please let me know if there are any other suggestions/comments I should take a look at, or if all is good now. Thanks!

locka99:
I'll take a look tomorrow. I've been away from the project for a few weeks so I have to catch up. Only pushed one change a few days ago.

locka99:
Thanks I've taken this in. I think when I look at the tokio 1.0 changes there might be a further refactor to consolidate the performance tuning settings in their own config section.

svanharmelen:
Thanks! And sounds good 👍🏻

PR #83, opened 2020-12-08
client support multi chunks

nkbai:
client support multi chunks:
and it's better to move MessageChunkWithChunkInfo to core module,but that will cause refactor of chunk


locka99:
Thanks for the patch I'll take it because I've tested the patch and it handles the CreateSessionResponse but there are still issues with the server.

PR #84, opened 2020-12-16
Add option to skip additional certificate validations

svanharmelen:
This option is sometimes also called `ssl-no-verify` and allows clients to accept certificates that have incorrect values.

We need to connect to a lot of endpoints that we cannot manage ourselves and we get all kinds of errors with the certificates. Fixing those errors is difficult as we do not always have the required access ourselves, yet the endpoints are all 100% trusted. So this options allows us to connect dispite of those errors.

### DIFF
--- a/client/src/builder.rs
+++ b/client/src/builder.rs
@@ -157,6 +157,15 @@ impl ClientBuilder {
         self
     }
 
+    /// Sets whether the client should verify server certificates. Regardless of this setting,
+    /// server certificates are always checked to see if they are trusted and have a valid key
+    /// length. In addition (if `verify_server_certs` is unset or is set to `true`) it will
+    /// verify the hostname, application uri and the not before / after values to ensure validity.
+    pub fn verify_server_certs(mut self, verify_server_certs: bool) -> Self {
+        self.config.verify_server_certs = verify_server_certs;
+        self
+    }
+
     /// Sets the pki directory where client's own key pair is stored and where `/trusted` and
     /// `/rejected` server certificates are stored.
     pub fn pki_dir<T>(mut self, pki_dir: T) -> Self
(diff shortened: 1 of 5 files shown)


locka99:
skip_verify_certs should be false by default - security by default. Also worth checking if rejected certs are implicitly trusted if they're in the trusted/ folder without a patch

svanharmelen:
> skip_verify_certs should be false by default

Oeps, that's a refactor error from my side. I was definitely not intending to change the default behavior!

svanharmelen:
@locka99 just updated the default, sorry about that!

Not sure what your mean with `Also worth checking if rejected certs are implicitly trusted if they're in the trusted/ folder without a patch`?

So manually place an otherwise rejected cert in the trusted folder and then check if it will be accepted with `skip_verify_certs` set to `true`?

locka99:
To clarify my comment the current code puts certs in a rejected folder if it doesn't like them, but it may be if you just copy the rejected cert over to the trusted folder that it will function without adding any additional code. If it does then that might be the easiest thing for your use case.

svanharmelen:
Ah, I see... Well even if that would work, I'm afraid it would not be a solution for our use case. Our application offers end users the ability to add new OPC connections through a web interface. Of course they will not have access to the machine where this code actually runs on, so they will not (and should not) be able to manually move certificates around.

Regardless of our use case, I doubt if it would even work looking at how and when this code is called. From a client perspective this is called during the [`create_session`](https://github.com/locka99/opcua/blob/master/client/src/session.rs#L1002) call and is used to validate the certificate. Looking at the logic in the [`validate_or_reject_application_instance_cert`](validate_application_instance_cert) function, it starts by making a call to [`validate_application_instance_cert`](https://github.com/locka99/opcua/blob/master/crypto/src/certificate_store.rs#L246) which then does all tests top to bottom. So it does not return early when the cert is confirmed to be in the trusted dir...

locka99:
Thanks I'll take it if it solves a problem for you. I may need to come back and revise some of the security settings from a config point of view because I have some work to do them for certificate signing / revocation I'll have to do at some point.

svanharmelen:
Thanks @locka99! Really glad this could be merged as it really does solve a problem for us. Thanks!

PR #90, opened 2021-01-25
Always use the supplied hostname when connecting

svanharmelen:
When you use the function `connect_to_endpoint_id` the code uses a function called `find_matching_endpoint` which substitutes the advertised hostname with the one the client supplied.

But when using `connect_to_endpoint` it uses a function called `find_server_endpoint` which does successfully match even if the hostname is different, but in that case it doesn’t substitute the advertised hostname.

This can cause issues when using IP addresses i.c.w. a NAT connection, have a difference in the actual hostname vs the used DNS name or when the server uses a hostname that is not resolvable so you need to connect using an IP address.

It seems this was also an issue with `connect_to_endpoint_id` as there is a mention of this in the code:

```
// Issue #16, #17 - the server may advertise an endpoint whose hostname is inaccessible
// to the client so substitute the advertised hostname with the one the client supplied.
```

So I figured we could just update the code to use that same code path in `connect_to_endpoint` as well. Hope this is indeed OK and that there isn’t a specific reason for not allowing this when using `connect_to_endpoint`.

Thanks!

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -13,7 +13,7 @@ use std::{
 use opcua_core::{
     comms::url::{
         hostname_from_url, is_opc_ua_binary_url, is_valid_opc_ua_url, server_url_from_endpoint_url,
-        url_matches, url_matches_except_host, url_with_replaced_hostname,
+        url_matches_except_host, url_with_replaced_hostname,
     },
     config::Config,
 };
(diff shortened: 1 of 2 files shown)


svanharmelen:
@locka99 any feedback on this one?

locka99:
Thanks I'll take this. Sorry for the delay.

svanharmelen:
Thanks!

PR #102, opened 2021-03-31
Make sure to register the new subscription

svanharmelen:
This seems to have been lost in the last changes, so adding it back to make `master` work again 😉


svanharmelen:
FYI, besides this missing bit things seem to work pretty smooth now. Also a lot less messages being send between the client and the server, so looks really good 👍🏻

PR #103, opened 2021-03-31
Allow the client to use the server time

### LINKED ISSUE #89: BadSecurityChecksFailed when checking signature
I'm using the client to connect to a 3rd party OPC UA server and for a couple of connections we get a lot of these kinds of errors:

```
[2021-01-21T20:34:35.905Z][INFO]-[ERROR opcua_client::comms::tcp_transport] Expecting a chunk, got an error message BadSecurityChecksFailed
[2021-01-21T20:34:35.905Z][INFO]-[ERROR opcua_client::comms::tcp_transport] Reader is putting connection into a finished state with status BadSecurityChecksFailed
[2021-01-21T20:34:35.905Z][INFO]-[DEBUG opcua_client::comms::tcp_transport] Reader is sending a quit to the writer
[2021-01-21T20:34:35.905Z][INFO]-[ERROR opcua_client::comms::tcp_transport] Read loop error Kind(ConnectionReset)
[2021-01-21T20:34:35.905Z][INFO]-[INFO  opcua_client::comms::tcp_transport] ReadState has dropped
[2021-01-21T20:34:35.905Z][INFO]-[DEBUG opcua_client::comms::tcp_transport] Read loop ended with an error
```

A lot means every 30 to 60 seconds (give or take). In between a couple of hundred measurements are successfully received but after some time it gets this error and then reconnects.

Any idea what could be the problem or how I can troubleshoot the issue?

Thanks!

svanharmelen:
In some situations, were the server setup is managed by a 3rd party, it can be needed to “follow” the server time in the client in order to be able to setup a connection.

This solution is pretty much inline with what is documented [here](https://reference.opcfoundation.org/v104/Core/docs/Part6/6.3/), except in this solution only the time used in the client is adjusted and the host clock is left untouched.

And just to be clear, the offset is only used for the communication and security related parts between the client and the server. For the rest the actuall time will still be used.

I tested this against several 3rd party server and the change resulted in stable connections without any errors, reconnects or constantly renewing security channels.

Fixes #89

### DIFF
--- a/client/src/session_retry.rs
+++ b/client/src/session_retry.rs
@@ -2,8 +2,9 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (C) 2017-2020 Adam Lock
 
-use chrono::{DateTime, TimeZone, Utc};
-use time::Duration;
+use chrono::Duration;
+
+use opcua_types::date_time::DateTime;
 
 #[derive(PartialEq, Debug)]
 pub enum Answer {
@@ -38,7 +39,7 @@ pub struct SessionRetryPolicy {
     /// this value is reset.
     retry_count: u32,
     /// The last retry attempt timestamp.
-    last_attempt: DateTime<Utc>,
+    last_attempt: DateTime,
 }
 
 impl Default for SessionRetryPolicy {
@@ -108,8 +109,8 @@ impl SessionRetryPolicy {
         Self::new(session_timeout, 0, 0)
     }
 
-    fn last_attempt_default() -> DateTime<Utc> {
-        Utc.ymd(1900, 1, 1).and_hms(0, 0, 0)
+    fn last_attempt_default() -> DateTime {
+        DateTime::ymd(1900, 1, 1)
     }
 
     pub fn session_timeout(&self) -> f64 {
@@ -128,13 +129,13 @@ impl SessionRetryPolicy {
         self.retry_count = 0;
     }
 
-    pub fn set_last_attempt(&mut self, last_attempt: DateTime<Utc>) {
+    pub fn set_last_attempt(&mut self, last_attempt: DateTime) {
         self.last_attempt = last_attempt;
     }
 
     /// Asks the policy, given the last retry attempt, should we try to connect again, wait a period of time
     /// or give up entirely.
-    pub fn should_retry_connect(&self, now: DateTime<Utc>) -> Answer {
+    pub fn should_retry_connect(&self, now: DateTime) -> Answer {
         if let Some(retry_limit) = self.retry_limit {
             if self.retry_count >= retry_limit {
                 // Number of retries have been exceeded
@@ -164,7 +165,7 @@ impl SessionRetryPolicy {
 fn session_retry() {
     let mut session_retry = SessionRetryPolicy::default();
 
-    let now = Utc::now();
+    let now = DateTime::now();
 
     let retry_interval =
         Duration::milliseconds(SessionRetryPolicy::DEFAULT_RETRY_INTERVAL_MS as i64);
@@ -194,7 +195,7 @@ fn session_retry() {
 #[test]
 fn session_retry_infinity() {
     let session_retry = SessionRetryPolicy::infinity(444.444, 1000);
-    let now = Utc::now();
+    let now = DateTime::now();
     assert_eq!(session_retry.should_retry_connect(now), Answer::Retry);
     assert_eq!(session_retry.session_timeout(), 444.444);
 }
@@ -202,7 +203,7 @@ fn session_retry_infinity() {
 #[test]
 fn session_retry_never() {
     let session_retry = SessionRetryPolicy::never(987.123);
-    let now = Utc::now();
+    let now = DateTime::now();
     assert_eq!(session_retry.should_retry_connect(now), Answer::GiveUp);
     assert_eq!(session_retry.session_timeout(), 987.123);
 }
--- a/client/src/session_state.rs
+++ b/client/src/session_state.rs
@@ -3,7 +3,6 @@
 // Copyright (C) 2017-2020 Adam Lock
 
 use std::{
-    self,
     sync::{
         atomic::{AtomicU32, Ordering},
         Arc, RwLock,
@@ -11,7 +10,7 @@ use std::{
     u32,
 };
 
-use chrono;
+use chrono::Duration;
 
 use opcua_core::{
     comms::secure_channel::SecureChannel, handle::Handle, supported_message::SupportedMessage,
@@ -49,6 +48,10 @@ lazy_static! {
 pub(crate) struct SessionState {
     /// A unique identifier for the session, this is NOT the session id assigned after a session is created
     id: u32,
+    /// Time offset between the client and the server.
+    client_offset: Duration,
+    /// Ignore clock skew between the client and the server.
+    ignore_clock_skew: bool,
     /// Secure channel information
     secure_channel: Arc<RwLock<SecureChannel>>,
     /// Connection state - what the session's connection is currently doing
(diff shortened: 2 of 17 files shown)


svanharmelen:
I thought of another solution for this problem that I want to try out... I expect it will make the diff a bit bigger, but I think it will be a more robust solution that solves an additional problem.

Will let you know if it works out and when the PR is ready to be reviewed.

locka99:
okay thanks - I did glance over the first patch and realised it would take some concentration to see what it was doing

svanharmelen:
@locka99 I think it's a good time to get your input on this one...

I kept the initial commit and added two new commits. The first one is a very big commit, but it essentially only does a rename. So probably best to focus on the other two commits.

The initial commit introduces a solution to allow the client to connect to a server who's clock isn't synced. On itself the solution works perfectly and allows you to connect without continuous reconnects or other issues. But it only focuses on the session itself, not on the actual data being send.

So the last commit updates the solution so it applies the offset to _all_ timestamps that are decoded by the client (except for the `source timestamp` for obvious reasons). The big advantage is that it means that all data will use the correct time as known by the client. So this assumes the client is configured correctly and that the server is outside of your control.

Now I did look at renaming `DecodingOptions` to `EncoderOptions` and also passing it to all `encode` methods. But that turned out to be an even bigger change and I wasn't sure if that was actually needed (I don't know exactly what timestamps the client sends to the server except for the timestamp in the headers).

So while it "feels" like an even more robust solution (the offset is then applied both ways during encoding and decoding which means all timestamps are consistent and you don't have to think about any exceptions in the code), I wasn't sure if it was worth the extra code and logic needed to pass the `EncoderOptions` around to every location that calls the `encode` method. I also looked at making the `encode` method accept `Option<EncoderOptions>` to make it easier, but that was also the point that I thought to first discuss all this with you 😏 

Very open to add the missing logic to make the move to `EncoderOptions` if you want to go that route, but curious what you think of all this and if the current solution is maybe enough to support the ability for clients to be able to work with servers with time drift.

Also open for any kind of video call if that would make this conversation easier. Just let me know if you prefer that 😉

locka99:
Would you be able to rename DecodingOptions back to DecodingLimits and decoding_options to decoding_limits? I think it would shrink the size of the diff down and make it easier what has changed. We can do that separately if you think it's necessary.

svanharmelen:
Yeah I can revert that, give me 10 minutes 😉

svanharmelen:
Make that 25 minutes... Got distracted a bit 😉 But I refactored things a bit so it only shows the functional changes. I do have the other commits stored in another branch, so maybe we want to revert it back after your review? Of indeed make a dedicated PR for it once this one has landed...

client/src/message_queue.rs
```diff
@@ -85,6 +85,7 @@ impl MessageQueue {
     pub(crate) fn store_response(&mut self, response: SupportedMessage) {
         // Remove corresponding request handle from inflight queue, add to responses
         let request_handle = response.request_handle();
+        trace!("Received response {:?}", response);
```
locka99, line 88: Remove this

client/src/session_retry.rs
```diff
@@ -108,8 +109,8 @@ impl SessionRetryPolicy {
         Self::new(session_timeout, 0, 0)
     }
 
-    fn last_attempt_default() -> DateTime<Utc> {
-        Utc.ymd(1900, 1, 1).and_hms(0, 0, 0)
+    fn last_attempt_default() -> DateTime {
+        DateTime::ymd(1900, 1, 1)
```
locka99, line 113: Just make sure this also sets clock to 00:00:00

client/src/session_state.rs
```diff
@@ -49,6 +48,10 @@ lazy_static! {
 pub(crate) struct SessionState {
     /// A unique identifier for the session, this is NOT the session id assigned after a session is created
     id: u32,
+    /// Time offset between the client and the server.
+    client_offset: Duration,
```
locka99, line 52: Obviously there is still potential for time to drift during the session

client/src/session_state.rs
```diff
@@ -401,9 +407,16 @@ impl SessionState {
         };
         let response = self.send_request(request)?;
         if let SupportedMessage::OpenSecureChannelResponse(response) = response {
+            if self.ignore_clock_skew {
```
locka99, line 410: Can you add a comment here to say what this is doing for clarity, i.e. to compensate for the difference in time between client and server

samples/client.conf
```diff
@@ -41,6 +41,7 @@ endpoints:
 session_retry_limit: 10
 session_retry_interval: 10000
 session_timeout: 0
+ignore_clock_skew: false
```
locka99, line 44: Perhaps this should go on the performance section

types/src/encoding.rs
```diff
@@ -27,6 +28,8 @@ pub struct DecodingLimits {
     pub max_byte_string_length: usize,
     /// Maximum number of array elements. 0 actually means 0, i.e. no array permitted
     pub max_array_length: usize,
+    /// Time offset between the client and the server
```
locka99, line 31: Can you rename offset something a little more clear and document it only needs to be used from the client side

client/src/session_state.rs
svanharmelen: True... Yet it updates the offset every time the secure channel is (re)created or renewed.

client/src/message_queue.rs
svanharmelen: Done (will push an additional commit with all requested changes in a bit)

client/src/session_retry.rs
svanharmelen: Yes, I double checked and this is indeed the case. See how the time is actually set [here](https://github.com/svanharmelen/opcua/blob/feature/time-keeping/types/src/date_time.rs#L247).

samples/client.conf
svanharmelen: I don't have a strong preference, so if you prefer it to be added to the performance section I'm good with that. Shall I update that?

types/src/encoding.rs
svanharmelen: We can call it `client_offset` here as well? And then add a line to the comment as well. Would that work?

client/src/session_state.rs
svanharmelen: Done (will push an additional commit with all requested changes in a bit)

types/src/encoding.rs
svanharmelen: Went ahead and added a commit that renames the field to `client_offset` and extends the comment.

samples/client.conf
svanharmelen: Went ahead and added a commit to move the field in the performance section...

svanharmelen:
@locka99 I think (hope) that I addressed all your feedback. I did it in separate commits so it's easy to check, but I also retrofitted the changes into the original 3 commits that also do the rename.

I personally prefer those 3 updated commits as it also updates `limits` to `options` which seems to be a better name moving forward. So if/when you are OK with the changes, I can update this PR with those 3 updated commits and mark the PR as ready. But of course I can also just squash the current commits in this PR and use that single commit. Just let me know what you prefer...

locka99:
I'll take the limits => options thing separately. It was just the difference between hundreds of files changing vs 17. I'll approve this part and many thanks. I hadn't thought of server / client out of sync and if there is a fix that reconciles the two then maybe it can even become the default at some point.

locka99:
You'll have to take it out of draft for me to merge it tho

svanharmelen:
Cool 😃 I squashed the existing commits on this branch to clean things up and took the PR out of draft... So all yours 👍🏻 

I'll create a separate PR for the limits => options change tomorrow... Thanks!

svanharmelen:
Thanks for merging this one @locka99! It runs in a few locations where we had issues for a few days now and improved things a lot 👍🏻 

And just created the PR for the rename the `DecodingLimits` struct (#107) to finish things off 😉

Issue #109, opened 2021-05-24
Enumerations changed


schroeder-:
First of thanks for this great library. Its easy to get started and get thinks working.

I am building an OPCUA Pubsub library based on this crate. While doing so i noticed something with enumerations. I think in the 1.03 OPCUA specs enums should be encoded as i32 and should be single values. With the 1.04 specs the length of an enum is specified in opc.ua.types.xsd as "LengthInBits". It's not possible to encode/decode a FieldMetaData because it contains DataSetFieldFlags. 

Also some enums now can be bitflags, they are marked with "IsOptionSet". They should be implemente as bitflags! instead of enum.

example:
```
  <opc:EnumeratedType Name="DataSetFieldFlags" LengthInBits="16" IsOptionSet="true">
    <opc:EnumeratedValue Name="None" Value="0" />
    <opc:EnumeratedValue Name="PromotedField" Value="1" />
  </opc:EnumeratedType>
```

locka99:
I'll leave this open to take a look at it. Up until now all the enums have been the same length so it hasn't required any special code.

PR #110, opened 2021-05-24
Fix enumeration size and flags enumerations

schroeder-:
See #109

Needed to upgrade to a new nodeset version, which contained the needed data in opc.ua.types.bsd.


locka99:
Thanks I'll merge this in.

PR #111, opened 2021-06-15
Fix username/password authentication

laumann:
This fixes two issues that combined caused the client to be unable to connect to our OPC-UA server (Kepware).

Here's a snippet from the GetEndpointsResponse that details the endpoint configuration
```
	0000000000:             messageSecurityMode: None 
	0000000000:             securityPolicyUri: http://opcfoundation.org/UA/SecurityPolicy#None 
	0000000000:             userIdentityTokens []: Size: 2 
	0000000000:                userIdentityTokens [ 0 ]: 
	0000000000:                   policyId: UserName 
	0000000000:                   tokenType: 1 
	0000000000:                   issuedTokenType: [empty] 
	0000000000:                   isssuerEndpointUrl: [empty] 
	0000000000:                   securityPolicyUri: http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15 
```

Overall, the two problems were:

 - The selected security policy for authentication was `None`, when it should be `Basic128Rsa15`
 - The server's nonce (remote nonce) was ignored 

## crypto/user_identity: Fix selection of security policy 

In `make_user_name_identity_token()` the security policy to use when
authenticating with username and password is determined from the
user_token_policy and token_security_policy.

In our tests, if the server provides eg. Basic128Rsa15 as the
user token policy, the logic would select None as the policy and
subsequently not encrypt the password and fail to connect.

Inverting some of the logic appears to work (at least for our tests).

## core/comms/secure_channel: Always set remote nonce

The set_remote_nonce_from_byte_string() performs some checks against the
message security mode.

In our tests, we use security policy = None and security mode = None,
but the server sets Basic126Rsa15 for the user identity token policy
meaning that we still need to use the server nonce for authentication.

This patch removes all the checking around security policy and mode,
which allows the client to connect to our OPC-UA server (Kepware)
successfully.

### DIFF
--- a/core/src/comms/secure_channel.rs
+++ b/core/src/comms/secure_channel.rs
@@ -283,22 +283,12 @@ impl SecureChannel {
 
     /// Set their nonce which should be the same as the symmetric key
     pub fn set_remote_nonce_from_byte_string(&mut self, remote_nonce: &ByteString) -> Result<(), StatusCode> {
-        if self.security_policy != SecurityPolicy::None && (self.security_mode == MessageSecurityMode::Sign || self.security_mode == MessageSecurityMode::SignAndEncrypt) {
-            if let Some(ref remote_nonce) = remote_nonce.value {
-                if remote_nonce.len() != self.security_policy.secure_channel_nonce_length() {
-                    error!("Remote nonce is invalid length {}, expecting {}. {:?}", remote_nonce.len(), self.security_policy.secure_channel_nonce_length(), remote_nonce);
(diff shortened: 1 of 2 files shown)


laumann:
Hi @locka99 sorry for the noise, I wanted to open this PR against [omnioiot/opcua](/omnioiot/opcua) and do a full verification that it works before opening a PR for you. I'm also not convinced that the fix I've implemented is the right way to do it, but I think it works for our purposes for now.

Issue #58, opened 2020-07-09
User/Password authentification fails on SecurityPolicy#None


execuc:
Hello,

I use your library, which is very good and reliable, as a client to connect to a opcua server. 
Server endpoint is described as :

```yaml
EndpointDescription
    EndpointUrl: opc.tcp://a.b.c.d:12345
    Server: ApplicationDescription
[...]
    MessageSecurityMode: None (0x00000001)
    SecurityPolicyUri: http://opcfoundation.org/UA/SecurityPolicy#None
    UserIdentityTokens: Array of UserTokenPolicy
        ArraySize: 1
        [0]: UserTokenPolicy
            PolicyId: UserName
            UserTokenType: UserName (0x00000001)
            IssuedTokenType: [OpcUa Null String]
            IssuerEndpointUrl: [OpcUa Null String]
            SecurityPolicyUri: http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15
    TransportProfileUri: http://opcfoundation.org/UA-Profile/Transport/uatcp-uasc-uabinary
    SecurityLevel: 16
```

So there is no signature or encryption for message but password token has to be sent in RsaPkcs because of token Basic128Rsa15. But authentification is rejected by the server with a BadIdentityTokenInvalid error at ActivateSessionRequest state.

Token policy
------------

My first problem is I have to force the use of Basic128Rsa15 policy in `make_user_name_identity_token` function because the library seems to take the endpoint policy (None) instead of token policy resulting a clear password.

Is this a configuration problem on my side ?

Token content
-------------

My second problem is that authentication fails with the same error. In this case, the content of the token is incomplete before it is encrypted in `legacy_password_encrypt`. I noticed that the server_nonce is empty. In fact it is not set in `set_remote_nonce_from_byte_string` function because of endpoint None policy.

Is it a problem of endpoint/token mode of the server which is not possible in the opcua specifications or it is not managed by the library.

Thank you for your work

laumann:
Hi @execuc I want to jump in on this issue, as I think we've been grappling with the same problem. I opened #111 (by mistake), but I wanted to follow up here.

Our setup is basically identical to the one you describe: Endpoint security policy = None, message security mode = None, but the user token policy = Basic128rsa15. We've worked from the 0.8 tag exclusively.

On the token policy: We figure that the logic around determining the security policy is wrong, but we could be mistaken. It doesn't appear to be a configuration issue.

[...]

PR #112, opened 2021-06-17
crypto: Properly select encryption algo from security policies

laumann:
When calling the ActivateSession Service, the service should accept
UserIdentityToken. For the UserNameIdentityToken, we need to provide a
password and it can be the case that the password needs to be encrypted
even though the SecureChannel's SecurityPolicy is set to None.

This commit adjusts the selection of the EncryptionAlgorithm as outlined
in Table 187 Opc Part 4. In general, we should prefer the
UserTokenIdentity.SecurityPolicy, expect for when it's not specified.

See locka99/opcua#58

### DIFF
--- a/crypto/src/tests/authentication.rs
+++ b/crypto/src/tests/authentication.rs
@@ -111,8 +111,10 @@ fn user_name_identity_token_encrypted() {
         &password,
     )
     .unwrap();
-    assert!(token.encryption_algorithm.is_null());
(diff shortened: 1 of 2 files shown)


laumann:
@locka99 This is one part of the fix. The implementation is based on Table 187 Opc Part 4:

image

locka99 (approved):

locka99:
Thanks I'll take it !

Issue #113, opened 2021-06-22
Connecting to a server with certificate RSA (4096 Bits) 


schroeder-:
I try to connect to a server (Softing OPC UA Toolkit) with a certificate with public key length of 4096 Bits, it always returns BadInvalidArgrument when OpenSecureChannelRequest is send from client.

If I change the certficate to a certificate with RSA Keylength of 2048, everything works. 
Tried following security polices: Basic256Sha256, Aes256Sha256RsaPss, Aes128Sha256RsaOaep all have the same problem.

locka99:
It might be a chunk size issue, or an issue with the keylength itself. 

The code does enforce key sizes in security_policy.rs, is_valid_keylength() but only according to the spec, so Basic256Sha256, Aes256_Sha256_RsaPss and Aes128_Sha256_RsaOaep should accept key lengths between 2048 & 4096.

The other option is perhaps there is some kind of chunking issue on the server. 

If you have logging enabled (e.g. setting RUST_OPCUA_LOG=debug when you have the console logger setup), then you should capture the debug level to see if you can isolate if the client is complaining about the handshake for a more specific reason than BadInvalidArgument.

schroeder-:
I attached the log, but the server only response with BadInvalidArgument. If I have time I will try to test it against an open62451 server with a rsa 4096 cert. 

Here is the log:
```
05:45:07 [WARN] Endpoint config contains no endpoints
05:45:07 [INFO] Connect
05:45:07 [INFO] Security policy = None
05:45:07 [INFO] Security mode = None
05:45:07 [DEBUG] (1) opcua_client::comms::tcp_transport: Creating a connection task to connect to [::1]:4840 with url opc.tcp://localhost:4840/
05:45:07 [DEBUG] (1) opcua_core::runtime: registering component connection-task, 1
05:45:07 [DEBUG] (1) opcua_client::comms::tcp_transport: Waiting for a connect (or failure to connect)
05:45:07 [DEBUG] (2) opcua_client::comms::tcp_transport: Client tokio tasks are starting for connection
05:45:07 [DEBUG] (2) opcua_core::runtime: registering component client-connection-thread-ThreadId(2)
05:45:07 [DEBUG] (3) opcua_client::comms::tcp_transport: Sending HELLO
05:45:07 [DEBUG] (3) opcua_core::runtime: registering component finished-monitor-task, 1
05:45:07 [DEBUG] (3) opcua_core::runtime: registering component read-task, 1
[...]

PR #116, opened 2021-06-24
Async Padding use remote certificate keylength

schroeder-:
For the asymmetric padding the server certificate length, is used to send the extra padding byte.
Fixes: #113

I need to run some tests, to ensure all cases are handled correct.


locka99:
Once tests pass I'll accept it - pending at the moment.

locka99:
Thanks a mil for the fix - hopefully it resolves the connection issues for others.

PR #117, opened 2021-06-24
Fix the rustfmt skip logic as it breaks with Rust 1.53

svanharmelen:
Instead of inserting `#[rustfmt::skip]` the files that need to be excluded are now described in the `rustfmt.toml` file.

https://github.com/rust-lang/rust/issues/64266


locka99:
Thanks I'll take this - much more elegant!

PR #119, opened 2021-06-30
Add a log trace that shows the response

svanharmelen:
I use this during debugging, so would be nice if this can be added...


locka99:
Thanks I'll add this although Wireshark is also a very useful tool for debugging traffic

PR #136, opened 2021-09-22
decode null array

schroeder-:
Make it possible to decode a null array. See [Part6 5.2.5](https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.5/)


locka99:
Thanks I'll take it

PR #138, opened 2021-09-29
feat: use CloseSession request when disconnecting

milgner:
When using the library - excellent work, btw! :D - I noticed that the sessions would remain open in the OPC UA servers I was connecting to.

Since the current code was already deleting all subscriptions anyway, I decided to also add the [`CloseSession`](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.6.4/) request.

One thing I'm not sure about is how to best test this - if you can give me a hint, I'll gladly do something about that.

### DIFF
--- a/client/src/session/session.rs
+++ b/client/src/session/session.rs
@@ -569,7 +569,7 @@ impl Session {
     /// or retrieve any existing subscriptions.
     pub fn disconnect(&self) {
         if self.is_connected() {
-            let _ = self.delete_all_subscriptions();
+            let _ = self.close_session_and_delete_subscriptions();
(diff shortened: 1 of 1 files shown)


saabye-io:
I have just merged this into master on on my own fork https://github.com/saabye-io/opcua and tested it with "OPC UA Server Simulation" from Integration Objects, which shows all sessions open sessions. 

This PR works perfectly and sessions are gone when I disconnect. 

We really need to be able to close the sessions as small PLCs does not always handle this well.

locka99:
Thanks I'll take it.

PR #144, opened 2021-11-24
fix docs typo

kimim:
I'm going through the nice documents, and find some typos to fix.


locka99:
I've added these in and thanks!

PR #151, opened 2021-12-28
Refactor: zero length variant array decoding

### LINKED ISSUE #145: Support Variant array_length == 0
Although I personally would prefer using `Variant::Empty` in case of an empty array, I yesterday encountered [this error](https://github.com/locka99/opcua/blob/4527d410782672fa731e526f922970448f084a93/types/src/variant.rs#L726) when our client tried to decode the response from a server that sent a zero-length array.

Now this code does look very explicit about its intent and I was wondering whether there's anything in the OPC UA spec that forbids arrays of length 0?

In https://open62541.org/doc/0.3/types.html#variant they seem to have many different variations on dealing with array length 0.

milgner:
Closes #145 

As I was just on the train with a few hours to kill, I took another look at this issue.

In the discussion, @locka99 pointed out that there were problems inferring the type of 0-length arrays. Yet after reviewing  https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.2/ it seems to me that, if the array bits are set, the type id should be there as well, even if the array length is 0. (the code being executed is in a branch where the array bit is set, so the encoding mask as a whole is not 0)
So I think the most likely explanation would be for an encoder implementation having gone awry on that part. If this happens now, `from_encoding_mask` will return `VariantTypeId::Empty`, which, in my opinion, is okay.

At the very least, this feels much cleaner than my initial idea of returning `Variant::Empty` from the encoder - which would probably also require some kind `PartialEq` implementation that can deal with comparisons of `Variant::Empty` and zero-length arrays.

### DIFF
--- a/lib/src/types/variant.rs
+++ b/lib/src/types/variant.rs
@@ -772,14 +772,15 @@ impl BinaryEncoder<Variant> for Variant {
         // Read array length
         let array_length = if encoding_mask & EncodingMask::ARRAY_VALUES_BIT != 0 {
             let array_length = i32::decode(stream, decoding_options)?;
-            // null array of type
-            if array_length == -1 {
-                let value_type_id = VariantTypeId::from_encoding_mask(element_encoding_mask)?;
-                return Array::new_multi(value_type_id, Vec::new(), Vec::new()).map(Variant::from);
-            }
-            if array_length <= 0 {
+            if array_length < -1 {
                 error!("Invalid array_length {}", array_length);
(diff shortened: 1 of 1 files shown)


thedanielhanke:
any news on this?

milgner:
I'm gonna rebase this onto the current `master` branch and update the MR.

milgner:
Rebased; tests are green locally. Not sure why the Appveyor integration fails? :thinking:

locka99 (approved):

Issue #153, opened 2022-01-13
[Question] Generating structs from XML for clients


Pascal-So:
disclaimer: I'm not too familiar with the whole information modelling / nodeset2 workflow so please feel free to point out if you think I might be misunderstanding something.

Assuming I am given the nodeset2 xml, bsd, csv, etc. files, my understanding is that these files can be used to set up object types, object instances, and further nodes on the server, and can also be used to inform a client about these nodes.

Thanks to the [Readme for the schema tools](https://github.com/locka99/opcua/tree/master/tools/schema#nodeset-compiler) I was able to take a nodeset2.xml and generate a rust module with which a server address space can be populated.

My question is now, what would the corresponding step for clients look like? I found the `gen_types.js` which takes a `.bsd` file and generates rust structs along with `impl BinaryEncoder` which is what I want, but the script doesn't look like it's intended to be modified to generate other types than the default types defined by the OPC UA Spec.

If my goal is to interact with a server that uses namespaces for which I have the nodeset2 files (e.g. one of the EUROMAP standards), what's the intended way to get this information into my code, preferably in a way to have as much information as possible already present at compile time instead of having to get information from the server at runtime?

Btw thank you for working on this library, so far this looks very pleasant to use overall!

schroeder-:
Currently it's not possible to generate from .bsd files. To do this we need to modify gen_types.js to allow extern .bsd files and change the outputs. Maybe i can have a look into it.

locka99:
The nodeset schema files aren't really describing structs, they're describing nodes in the address space - objects, variables, data types, methods and the relationships (references) between each other, e.g. that a variable is a child of an object and so on. 

There are enums & structs in the .bsd files which are turned into generated types so they can be serialized into and out of extension objects. Most of the fundamental types are handwritten and only the types / enums that are composites of fundamental types are machine generated. I'm not sure if it is .bsd types you are after or the nodeset?

At present the node set compiler in the project is mostly about populating the address space at startup and doesn't really help much beyond that. I guess the code generators could be extended in useful ways, e.g. bespoke builders to assist servers to make instances of objects and so on.

From the client side, it might be possible to create helpers that allow someone to monitor all the vars of an object in one go (for example) but it wouldn't be a structure, more of just a helper to get all the stuff in one subscription. If you have an example of a commercial or open source OPC UA library that does generate code in a useful way I can take a look at it.

Pascal-So:
[...]

PR #154, opened 2022-01-15
schema: allow code generation from bsd

schroeder-:
Allow enum and struct generation from bsd files, for use supplied bsd files. 
fixes: #153


locka99:
Thanks I'll take the PR to close the issue #153

PR #169, opened 2022-02-24
Use tokio version "1"

laumann:
We use tokio 1.16 internally meaning that we cannot select a compatible version when using this library as a dependency.

I only want the version requirements to be a little more lax, not necessarily `"1"` but I thought this could get us started :-)

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -25,7 +25,7 @@ chrono = "0.4"
 serde = "1.0"
 serde_derive = "1.0"
 futures = "0.3"
-tokio = { version = "~1.15", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
 tokio-util = { version = "0.6.9", features = ["codec"] }
 lazy_static = "1.4.0"
 
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -28,7 +28,7 @@ serde_derive = "1.0"
 serde_yaml = "0.8"
 lazy_static = "1.4.0"
 futures = "0.3"
-tokio = { version = "~1.15", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
 tokio-util = { version = "0.6.9", features = ["codec"] }
 bytes = "1.0.1"
 url = "1.6"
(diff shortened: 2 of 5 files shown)


laumann:
An alternative is ">=1.15" I suppose if you just want to go for a minimum version.

svanharmelen:
I have the same issue and currently use a forked branch because of this. Would be very nice if this constraint could be relaxed just a little...

laumann:
@svanharmelen Good to know. What version of tokio do you use?

svanharmelen:
`1.17` at the moment...

locka99:
If I loosen it though, wouldn't that completely mess up the lock file?

laumann:
> If I loosen it though, wouldn't that completely mess up the lock file?

I don't think so. You made me wonder why there weren't any changes to Cargo.lock, but there shouldn't be any. As I understand Cargo.lock, it just records the version of a dependency that was chosen at some point, whereas the dependency specification in Cargo.toml puts bounds on the accepted versions of dependencies.

With this change, if you run `cargo update` you'd probably get the tokio versions bumped to 1.17 (or whatever's latest) in Cargo.lock, but until that point nothing should change.

svanharmelen:
Indeed it doesn't directly impact the lockfile in this repo (only when you update), but when using the client as a library you are not using the lockfile in this repo. Instead you will use the lockfile that is associated with your own binary.

locka99:
I've changed the deps to be 1 for tokio and 0.6 for tokio-util. This PR can therefore be closed. I think if it causes issues I'll have to be more specific but I'll leave it like this for now.

svanharmelen:
Nice 👍🏻

lovasoa:
Cool ! Could this be pushed to crates.io ? The latest version in the repo seems to still depend on tokio 0.1...

locka99:
I'm thinking of publishing 0.9 soon to get on with async client as 0.10 but I have to go through things to make sure there are no regressions.

laumann:
> I'm thinking of publishing 0.9 soon to get on with async client as 0.10 but I have to go through things to make sure there are no regressions.

Nice! Anything you need in that regard? I made a sample browse-client (to scan the address space of an OPC-UA server) - would you be interested in including that?

Issue #172, opened 2022-03-02
when a getter returns an error, the entire server crashes


lovasoa:
When a getter returns  a value such as `StatusCode::BadTypeMismatch`, the entire OPCUA server crashes.

```
 'tokio-runtime-worker-6' panicked at 'called `Result::unwrap()` on an `Err` value: IS_ERROR | BadUnexpectedError | BadResourceUnavailable | BadCommunicationError | BadEncodingLimitsExceeded | BadUnknownResponse | BadShutdown | BadServerNotConnected | BadTcpMessageTooLarge | BadTcpNotEnoughResources | BadRequestInterrupted | BadRequestTimeout | BadSequenceNumberInvalid | BadConfigurationError | BadSensorFailure | BadOutOfService', /home/olojkine/.cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.8.1/src/address_space/variable.rs:357:126
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::result::unwrap_failed
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1613:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1295:23
   4: opcua_server::address_space::variable::Variable::value
             at /home/olojkine/.cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.8.1/src/address_space/variable.rs:357:13
   5: <opcua_server::address_space::variable::Variable as opcua_server::address_space::node::Node>::get_attribute_max_age
             at /home/olojkine/.cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.8.1/src/address_space/variable.rs:172:40
   6: opcua_server::services::attribute::AttributeService::read_node_value
             at /home/olojkine/.cargo/registry/src/github.com-1ecc6299db9ec823/opcua-server-0.8.1/src/services/attribute.rs:363:49
   7: opcua_server::services::attribute::AttributeService::read::{{closure}}
[...]

PR #173, opened 2022-03-02
do not crash when a getter returns an error

lovasoa:
fixes https://github.com/locka99/opcua/issues/172


lovasoa:
Thanks for merging ! Could you publish a new release on crates.io so that I can use it in my project ?

Issue #182, opened 2022-03-24
invalid nonce when opening secure channel


lovasoa:
When trying to connect to a rust opcua server with a SecurityPolicy of None from ignition, one gets the following error:

```
    UaException: status=Bad_NonceInvalid, message=nonce must be non-zero
``` 

Indeed, it looks like the server always sends a null string instead of a random nonce

https://reference.opcfoundation.org/v104/Core/docs/Part4/5.5.2/

lovasoa:
Here is a packet capture of the exchange between the "example server" from this repo and an out-of-the-box ignition client:

[pcap.zip](https://github.com/locka99/opcua/files/8344437/pcap.zip)

Here is a screenshot of what ignition shows:

![image](https://user-images.githubusercontent.com/552629/159986833-ab26a78f-0557-4f5d-9c08-6c9daf8c47c4.png)

![image](https://user-images.githubusercontent.com/552629/159986996-169e9f7c-13e7-4309-8533-dd68b801a3dc.png)

lovasoa:
I tracked down this issue to https://github.com/locka99/opcua/commit/5c20ed8f1d3b708b5c0441528eea792a9607fa07

The server sends a nonce composed of 32 zeroes instead of the previous correct behavior of not including a nonce when the security policy is None.

locka99:
Yes I think this nonce stuff goes back and forth depending on the implementation being used against. The spec doesn't specify a nonce length and some client/server combinations want it and others do not

locka99:
I might revert it back to 0 but I'll have to review the change that put it to 32 bytes

lovasoa:
You can send a null  nonce to signal you don't have a nonce when the security policy is none, but sending an all-zero nonce is a big no no, in any situation.

locka99:
I meant a zero length nonce

PR #183, opened 2022-03-24
fix server connexions from ignition

lovasoa:
Fixes https://github.com/locka99/opcua/issues/182

Reverts a change introduced in 5c20ed8f1d3b708b5c0441528eea792a9607fa07

### DIFF
--- a/core/src/comms/secure_channel.rs
+++ b/core/src/comms/secure_channel.rs
@@ -1031,7 +1031,7 @@ impl SecureChannel {
 
     pub fn local_nonce_as_byte_string(&self) -> ByteString {
         if self.local_nonce.is_empty() {
-            ByteString::from(&[0u8; 32])
+            ByteString::null()
         } else {
             ByteString::from(&self.local_nonce)
         }
--- a/core/src/comms/secure_channel.rs
+++ b/core/src/comms/secure_channel.rs
@@ -12,14 +12,14 @@ use chrono::Duration;
 
 use opcua_crypto::{
     aeskey::AesKey,
+    CertificateStore,
     pkey::{KeySize, PrivateKey, PublicKey},
     random,
-    x509::X509,
-    CertificateStore, SecurityPolicy,
+    SecurityPolicy, x509::X509,
 };
 use opcua_types::{
-    service_types::ChannelSecurityToken, status_code::StatusCode, write_bytes, write_u8,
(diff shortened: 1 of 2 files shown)


svanharmelen:
Reverting the change like this (so in full) will reintroduce the problems that were solved by this change. So I suggest to look for a solution that solves this particular issue, instead of rolling back the whole change...

lovasoa:
@svanharmelen : Did you look at the pr ? This is not a simple git revert of the previous change. The previous change did not include any comment, but I suppose it introduced an all-zeroes nonce to force connections to clients that require a nonce during secure channel creation. This pr keeps the nonce, but makes it an actual random nonce, as required by the spec.

svanharmelen:
Looks like I might have mixed things up... I was looking at [this](https://github.com/locka99/opcua/pull/183/files#diff-c1177d127d0a14e056ad2ce44df9c5b63244b1d9889862a1a965c5f030aa89a1L275-L284) part which make me think of this change: https://github.com/locka99/opcua/pull/114/files

But it seems your change is not touching that code actually 👍🏻

lovasoa:
@locka99 : is there something you want me to do before this can be merged ? We depend on this for our server.

locka99:
I'll take it since it uses a zero length array instead of null but we'll to monitor if it makes any servers unhappy. & thanks for your persistence in this.

locka99:
I just noticed SecurityPolicy::None is given a 32 byte array. Can you check please if 0 works?

lovasoa:
I am no expert, but see https://github.com/eclipse/milo/pull/949 : 

>> is it a standard-compliant behavior to send a null nonce ?
>
> It is only compliant when the null nonce is in the OpenSecureChannelResponse and the channels are not using security.
>
> There are other nonce exchanges, in the CreateSessionResponse and ActivateSessionResponse, and in these it is important in regards to the CVE that the nonce be non-empty and non-zero if they are going to be used in the subsequent encryption of credentials for transit on an otherwise unsecure channel.

locka99:
There is a underlying difference between a null bytestring and a zero length bytestring. It may be that using 0 satisfies everyone instead of null. I'll have to see if the specs say because this stuff has been a constant bane - some server complains when there is a nonce and others complain when there is a null.

locka99:
I'll release a 0.9.1 with your changes cherry-picked into it

Issue #189, opened 2022-04-07
Failing to detect lost connection and reconnect


thosaa:
Hi 
I'm starting a client and is connected to a Microsoft OPC PLC server (mcr.microsoft.com/iotedge/opc-plc). 

The client connects, subscribes and get event notifications, however, I'm having an issue with detecting a lost connection and reconnect. 
I'm using the current master branch, but have seen the same behaviour in 0.8 and 0.9.

The client is started with a quite aggressive retry and time-out:
```rust
let client = Box::new(ClientBuilder::new()
        .application_name("CDGs OPC UA Client")
        .application_uri("urn:Discovery")
        .product_uri("urn:OpcuaClient")
        .create_sample_keypair(true)
        .trust_server_certs(true)
        .ignore_clock_skew()
        .session_retry_interval(5000)
        .session_timeout(5000)
        .session_retry_limit(-1)
        .client().unwrap());
```

And I have set up callback on connection state change:
```rust
session.set_connection_status_callback(ConnectionStatusCallback::new(move |connected| { 
      info!("**** Connection status is now: {} ****", connected); 
}));
```

When I close the OPC server with `docker stop {container_id}` the client stops getting events and starts with the keep-alive attempt.
This is what the log states:
```log
2022-04-07 13:38:35.783Z INFO  opcua_client::session::session - Session activity keep-alive request
2022-04-07 13:38:35.785Z DEBUG opcua_client::message_queue - Request 43 was processed by the server
2022-04-07 13:38:43.783Z INFO  opcua_client::session::session - Session activity keep-alive request
2022-04-07 13:38:43.785Z DEBUG opcua_client::message_queue - Request 44 was processed by the server
2022-04-07 13:38:43.785Z ERROR opcua_client::comms::tcp_transport - Write bytes task IO error Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
2022-04-07 13:38:51.783Z INFO  opcua_client::session::session - Session activity keep-alive request
2022-04-07 13:38:51.786Z DEBUG opcua_client::message_queue - Request 45 was processed by the server
2022-04-07 13:38:51.786Z ERROR opcua_client::comms::tcp_transport - Write bytes task IO error Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
2022-04-07 13:38:59.783Z INFO  opcua_client::session::session - Session activity keep-alive request
[...]

PR #196, opened 2022-04-22
[#189]: Fix detecting of failed connection

ctron:
The client failed to detect a broken connection, as it didn't react on
a write error, aside from logging it.

This change makes it set the connection to finished when a write error
occurs, which will trigger a re-connect.

However, due to the fact that in a single session actually two Tokio
runtimes are involved, the session drops a Tokio runtime from an async
context, which will panic.

In order to fix this, this PR also removes the explicit Tokio runtimes
from the TCP comms and session struct. This means a breaking change,
as now the client functions need to be called with an active Tokio
context. This is easy as using `#[tokio::main]` and
`Handle::spawn_blocking` for the synchronous functions, but might
involve some more work in other scenarios.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1569,6 +1569,7 @@ version = "0.10.0"
 dependencies = [
  "opcua",
  "pico-args",
+ "tokio 1.15.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


ctron:
Unfortunately there was no "0.10" branch, so this PR failed to merge with master.

svanharmelen:
@ctron maybe you can checkout this branch: https://github.com/svanharmelen/opcua/tree/feature/async-client it's intended to become a full async-client. I did run and test it a little, but not enough yet to create a PR for it to this repo...

svanharmelen:
FYI: #197

PR #203, opened 2022-05-06
Remove session from address space when terminated

lovasoa:


lovasoa:
This fixes a memory leak where a session object was not removed from the address space when the session ended

PR #213, opened 2022-06-02
correct binary encoding

schroeder-:
correct binary encoding for DiagnosticInfo

See: https://reference.opcfoundation.org/v104/Core/docs/Part4/7.8/


locka99:
I think the bits were correct - if you look at 5.2.2.12 in the specs they were in the proper order:

https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.2/

schroeder-:
You are right. Was confused because of https://reference.opcfoundation.org/v104/Core/docs/Part4/7.8/ where the order is different....

PR #217, opened 2022-06-20
Use DateTime for token creation time

### LINKED ISSUE #206: Secure channel lifetime calculated as difference between local and remote clock in client
When evaluating whether it is time to renew the secure channel token, the time that the current token has been used is calculated as a difference between the current time on the local computer and the time of creation reported by the server.

This means that if the clocks in the client and server are not in sync, the calculated elapsed lifetime is incorrect.

Note that I am not talking about time zones/local time, all times are handled correctly in UTC, the problem is just calculating the elapsed lifetime as a difference between current time on the local computer and the time of creation from the server.

This can be tested by connecting to an OPC UA server where the clock is a few hours before the clock on the machine where the opcua client is runnint.

Using the UaExpert client, I get a warning if the server time is off by more than 30 minutes compared with the local computer, but it still handles the channel lifetime correctly.

Jarngreipr:
SecureChannel.token_created_at was set using the OPC server time which could cause connection failures if the OPC server is not configured correctly.
Changed so that the same time source is used to determine the token lifetime (DateTime::now())
Fixes #206

### DIFF
--- a/lib/src/core/comms/secure_channel.rs
+++ b/lib/src/core/comms/secure_channel.rs
@@ -176,7 +176,7 @@ impl SecureChannel {
     pub fn set_security_token(&mut self, channel_token: ChannelSecurityToken) {
         self.secure_channel_id = channel_token.channel_id;
         self.token_id = channel_token.token_id;
-        self.token_created_at = channel_token.created_at;
+        self.token_created_at = DateTime::now();
         self.token_lifetime = channel_token.revised_lifetime;
     }
 


svanharmelen:
Hmm.. I know this is merged a long time ago, but I've been away from this project for a while and just noticed the changed line in the code while debugging something else. I think this approach is not the right way to solve your issue and may (in some corner cases) introduce unexpected behavior.

Settings the `created_at` like this (while it may solve your direct issue) would mean that the `created_at` time is no longer in sync with the rest of all the timestamps used throughout the code.

So I wonder if you have tried to enable the client feature called `ignore_clock_skew` as that was introduced to fix exactly this issue (https://github.com/locka99/opcua/pull/103)? I would suggest to turn on that feature and I will make a PR to revert this change so that all timestamps remain in sync and consistent.

Issue #150, opened 2021-12-21
UaException: Bad_NonceInvalid when connecting from inductiveautomation Ignition OPC Client


fautore:
Hi everybody,
I set up a opcua-demo-server as in `samples/demo-server` and used the default configuration, then i set up an Ignition server on the same machine and created a new OPC Client connection to localhost port 4855.

the endpoint i'm setting into ignition is the following:
`opc.tcp://localhost:4855/`
but the Ignition OPC Client keeps faulting with the following error message:
` UaException: status=Bad_NonceInvalid, message=nonce must be non-zero`

I also tried fiddling with the configuration of the demo-server, i tried dropping the endpoints that use encryption but i seem to have no luck with this.

I also set up a opcua-simple-client that i found on `samples/simple-client/`, default configuration and that seems to work fine.

Any ideas / help with the configuration i am running?
Any help will be appreciated,
Thank you.

schroeder-:
Which version are you using? Can you try current master? See issue #58 , which was fixed on master lately.

fautore:
Yes, downloaded master and compiled it yesterday.

schroeder-:
Maybe we have the same issue on the server side.
If you are using the configuration file try using this endpoint to force a password security policy:
```
 none:
    path: /
    security_policy: None
    security_mode: None
    security_level: 0
    password_security_policy: http://opcfoundation.org/UA/SecurityPolicy#Basic256
    user_token_ids:
      - ANONYMOUS
      - sample_password_user
      - sample_x509_user
```
Also does anonymous access with secruity_policy: None works?

fautore:
I tried just now with the configuration you suggested, same result, Ignition client side connection faults.

Also tried connecting from simple-client in localhost with "none" endpoint and it seems to work just fine.

milgner:
Not sure whether this is still current, but I encountered similar issues with the Prosys Simulation Server. The server was confused because the client sent a security certificate but no nonce. So instead of ignoring it for SecurityPolicy None, it complained.

There's a patch at https://github.com/re-gmbh/opcua/commit/2f6c80b9d0a1b7435f9b8d2661c5edf658e19cf9, I'll try to allocate a few minutes to create PR for it.

PR #228, opened 2022-09-14
Prevent sending client certificate for SecurityPolicy None

### LINKED ISSUE #225: Relax protocol check
Currently running version 0.10.0 against few commercially available server packages from
- Prosys
- UA Expert
- Integration Object

Connecting with UA Expert seems to work fine, but running against Prosys fails due to invalid Nonce error, and Integration Object due to BadCertificateUriInvalid error. Since both were working with earlier version of the library, I am guessing there has been some change in protocol handling/checks. 

I am wondering whether there are options to relax those checks when connection/sessions are created.

milgner:
Closes #225, #150

### DIFF
--- a/lib/src/client/session/session.rs
+++ b/lib/src/client/session/session.rs
@@ -1195,6 +1195,18 @@ impl Session {
             pass,
         )
     }
+
+    fn client_certificate_bytes(&self) -> ByteString {
+        if !self.security_policy().requires_certificate() {
+            return ByteString::null();
+        }
+        let certificate_store = trace_write_lock!(self.certificate_store);
+
+        match certificate_store.read_own_cert_and_pkey_optional() {
+            (Some(certificate), _) => certificate.as_byte_string(),
+            _ => ByteString::null()
+        }
+    }
 }
 
 impl Service for Session {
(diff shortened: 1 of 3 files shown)


AiyionPrime:
@milgner is there a section in the OPC UA spec which this PR could refer to regarding sending or omitting upon the lack of a SecurityPolicy?

AiyionPrime:
I can't find anything that forbids the client from sending the certificate.
A good approach might be [Security None CreateSession ActivateSession 1.0](https://profiles.opcfoundation.org/conformanceunit/2839) referenced in [Secourity Policy: None](https://profiles.opcfoundation.org/profile/762) though, which optionally suggests a fallback.

1. try connecting without a cert
2. if the server denies send it (like we do now)

I think merging this in the current state would not be wrong, but the current implementation is not either. If we're unlucky then we might just encounter the next person arguing to revoke this again.

I'd suggest implementing the fallback way, which should be the most compatible one, right?

milgner:
Sounds like a sensible approach. The specification is indeed not very concrete on the topic.

From a purely anecdotal perspective, we have been running this patch for a long time now and never encountered any server that wanted a certificate in combination with SecurityPolicy None.

Apart from that, it is unclear to me why it would want a certificate when there's no encryption to use it for. (Just like it doesn't make much sense to crash when there is one 😅)
But nonetheless it might be a good idea to have the fallback in case of quirky servers. 👍

PR #233, opened 2022-10-26
Implement error handling

matthiasbeyer:
This patch implements the first step to proper error handling instead of just logging errors using the `log` crate.

---

There's a lot to do in this crate to bring error handling and config validation up to the state of art. This is only the first step, basically "to get the foot into the door" for doing error handling the proper way here.

Let me know what you think!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1440,6 +1440,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "tempdir",
+ "thiserror",
  "time",
  "tokio 1.20.1",
  "tokio-util",
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -67,6 +67,7 @@ openssl-sys = "0.9"
 gethostname = "0.2.1"
 libc = "0.2"
 foreign-types = "0.3"
+thiserror = "1"
 
 [dependencies.env_logger]
 version = "0.9"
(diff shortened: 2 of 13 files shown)


locka99:
In principle I don't mind this for the configuration stuff but I could see that ConfigError bloating out to encompass other errors I already see a bit of it with the SecurityPolicy having to emit a ConfigError. It might be better for now that it doesn't and only the Config impls use that error type, or there is a mapping from a SecurityPolicyError into a ConfigError as it crosses that boundary.

matthiasbeyer:
So the end-goal would be to have one `Error` type for the whole crate and that might consist of several types, one for configuration, one for the security bits you talked about.. etc.

Pseudo code:


```rust
#[thiserror::Error]
pub enum Error {
    Config(#[from] ConfigError),
    Security(#[from] SecurityError),
}
```

And then let the crate only return that error type and let the user handle each error kind individually.

locka99:
I think for the time being confine it to configuration errors. That would be a well defined use case where we want the the client / server to just terminate and quit if config is not right. It might be that over time it extends elsewhere but generally logging is still desirable for most situations where it is possible to see a timeline of failures if there is a recoverable failure of some kind.

HerrMuellerluedenscheid:
I think this merge request should be reconsidered (and rebased). I think this is more idiomatic and I find that easier to digest when reading the code.

PR #240, opened 2022-11-18
fix out of order locking between server_state and session_manager 

joshuagleaton:
fix out of order locking between server_state and session_manager causing deadlock.

Deadlock occurred in the following scenario:

```
thread 1:
spawn_reading_loop_task:
	tcp_transport.rs:423:		transport.write_lock
handle_message():
	message_handler.rs:124		session_manager.write_lock
create_session()
	session.rs:45			server_state.write_lock

thread 2:
spawn_reading_loop_task
	tcp_transport.rs:423		transport.write_lock
close_session:
	session.rs:337			server_state.write_lock
	session.rs:345			session_manager.read_lock
```

### DIFF
--- a/lib/src/server/services/session.rs
+++ b/lib/src/server/services/session.rs
@@ -334,8 +334,6 @@ impl SessionService {
         address_space: Arc<RwLock<AddressSpace>>,
         request: &CloseSessionRequest,
     ) -> SupportedMessage {
-        let server_state = trace_write_lock!(server_state);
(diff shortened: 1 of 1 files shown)


locka99:
I'll take it although looking at server_state in the code it could probably even be a read-only lock in this scenario.

locka99:
Thanks for this - deadlocks can be a pain to id sometimes

PR #247, opened 2022-12-20
Add serde(default) to ServerConfig, implemented default for TcpConfig…

didu14:
Allows for defining partial configuration in file and take the rest from the default implementation.
TcpConfig and Performance receive their own default implementation, to be partially definable in the config too.


didu14:
Unfortunately, the history was destroyed after merge. I'll reopen the pull request with the changes only.

PR #255, opened 2023-02-16
Do not skip first subject alt name when validating hostname

gmalkas:
Hi there!

Thank you for this library.

I have recently run into an issue while writing a client using this library. The test server provided by an OEM (i.e., I can't change the server) has a certificate which includes two subject alternate names: one DNS and one application URI. The DNS hostname matches the one used to connect to the server, meaning the certificate is valid, but it comes first, so it is rejected by the library `is_hostname_valid` function.

The specification does not seem to mandate any specific order for subject alternative names:
https://reference.opcfoundation.org/Core/Part6/v105/docs/6.2.2

Skipping the first alt name can cause some valid server certificate to be rejected if the DNS name is the first one in the list.  I am unsure why the first subj alt name was skipped in the first place.

Thank you.

### DIFF
--- a/lib/src/crypto/x509.rs
+++ b/lib/src/crypto/x509.rs
@@ -451,10 +451,8 @@ impl X509 {
 
     fn subject_alt_names(&self) -> Option<Vec<String>> {
         if let Some(ref alt_names) = self.value.subject_alt_names() {
-            // Skip the application uri
             let subject_alt_names = alt_names
                 .iter()
-                .skip(1)
                 .map(|n| {
                     if let Some(dnsname) = n.dnsname() {
                         dnsname.to_string()
@@ -520,11 +518,9 @@ impl X509 {
             "is_application_uri_valid against {} on cert",
             application_uri
         );
-        // Expecting the first subject alternative name to be a uri that matches with the supplied
-        // application uri
         if let Some(ref alt_names) = self.value.subject_alt_names() {
-            if alt_names.len() > 0 {
-                if let Some(cert_application_uri) = alt_names[0].uri() {
+            if let Some(alt_name) = alt_names.iter().find(|alt_name| alt_name.uri().is_some()) {
+                if let Some(cert_application_uri) = alt_name.uri() {
                     if cert_application_uri == application_uri {
                         info!("Certificate application uri {} is good", application_uri);
                         StatusCode::Good
@@ -536,11 +532,10 @@ impl X509 {
                         StatusCode::BadCertificateUriInvalid
                     }
                 } else {
-                    error!("Cert's first subject alt name is not a uri and cannot be compared");
-                    StatusCode::BadCertificateUriInvalid
+                    unreachable!()
                 }
             } else {
-                error!("Cert has zero subject alt names");
+                error!("Cert has zero URI subject alt names");
                 StatusCode::BadCertificateUriInvalid
             }
         } else {


locka99:
My interpretation of the spec is what it says - applicationUri, hostnames - i.e., the first entry is the application uri and there are one or more hostnames / ip addresses thereafter. It's probably problematic to switch the ordering anyway since a uri / urn could potentially resemble an IPV6 address or vice versa. So even if the code were loosened it would still have to look for exactly one uri/urn somewhere in the list and filter it out.

Do you know how the client created its certificate and if it works with other OPC libraries? If it's a case of client error, then it's probably not a good idea to loosen security in the code for the sake of it.

gmalkas:
Thank you for taking a look.

In my case, I am writing the client, it is the server certificate that is problematic due to the order of subject alternative names. 

The open62541 C library looks at all SAN and picks the first one of type URI: https://github.com/open62541/open62541/blob/10047d4757eb2b37ca8d93c7124dcd421ce2ddbf/plugins/crypto/openssl/ua_pki_openssl.c#L634

For whatever it's worth, this library claims to be certified by the OPC Foundation: https://www.open62541.org/certification/

I am not super familiar with x509v3 but my reading of the openssl manpage indicates the subject alternative names are typed, meaning the ambiguity between URI/URN and IP wouldn't happen:

> Subject Alternative Name
> 
> This is a multi-valued extension that supports several types of name identifier, including email (an email address), URI (a uniform resource indicator), DNS (a DNS domain name), RID (a registered ID: OBJECT IDENTIFIER), IP (an IP address), dirName (a distinguished name), and otherName. The syntax of each is described in the following paragraphs.

(https://www.openssl.org/docs/manmaster/man5/x509v3_config.html)

Here is an example on a client certificate generated by UAExpert (https://www.unified-automation.com/products/development-tools/uaexpert.html):
```
X509v3 Subject Alternative Name:
                URI:urn:myhostname:UnifiedAutomation:UaExpert, DNS:myhostname
``` 

`URI:` and `DNS:` are indicating the type of the entry, so there is no ambiguity. If an IP was specified for the URI, it would still have a type of `URI`.

Does it make sense?

Thank you.

gmalkas:
Hi,

I've looked into this further and checked the official OPC Foundation reference implementation in C#.

They do verify the application URI/URN and server domain name by looking at the certificate subject alternative names of the relevant type.

In particular, they check the server domain name against every name of type DNS: https://github.com/OPCFoundation/UA-.NETStandard/blob/6d675b8cd0107372240fc530a95c5d2f59228820/Stack/Opc.Ua.Core/Security/Certificates/X509Utils.cs#L173 

And the application URI against the first name of type URI:
https://github.com/OPCFoundation/UA-.NETStandard/blob/7d9be41cf5337e7832819ef76f150b8573c1ef12/Libraries/Opc.Ua.Configuration/ApplicationInstance.cs#L636 and https://github.com/OPCFoundation/UA-.NETStandard/blob/6d675b8cd0107372240fc530a95c5d2f59228820/Stack/Opc.Ua.Core/Security/Certificates/X509Utils.cs#L127

I will refactor my change to match that implementation. Let me know if that makes sense.

Thanks!

locka99:
You might be able to replace the skip with a .filter(|n| n.dnsname().is_some() || n.ipaddress().is_some()) so it doesn't assume the application uri is first and will skip anything which is not a dnsname / ipaddress.

I think the code as-is is perhaps a little fragile and maybe needs work to enforce the spec more stringently. I don't have time to look at the other impls at the moment but think it could be changed to work with application uri(s) anywhere, favouring the first uri and filter out anything that is not a host/dnsnames from places that test that kind of thing.

PR #266, opened 2023-05-03
impl from Variant for primitives and prim. arrays, bool casting extended

divi255:
Good day,

great library, thanks! for my project it was missing casting from Variant to Rust primitives and arrays of primitives, submitting an implementation. I also extended casting from Variant:Boolean with casting to numeric ones.

### DIFF
--- a/lib/src/types/guid.rs
+++ b/lib/src/types/guid.rs
@@ -83,6 +83,12 @@ impl FromStr for Guid {
     }
 }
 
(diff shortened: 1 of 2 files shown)


locka99 (changes requested):
Other than the piece at the top, looks good

divi255:
done, sorry for the delay. removed endpoint config patch

locka99:
Thanks for the patch!

PR #277, opened 2023-07-04
{core,types}/fuzz: update dependencies

0x2ec:
Updates Cargo.toml files for the fuzz binaries, fixes up the fuzzer runners, and update Cargo.lock.

### DIFF
--- a/lib/src/core/fuzz/Cargo.toml
+++ b/lib/src/core/fuzz/Cargo.toml
@@ -11,20 +11,12 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4"
 bytes = "1.0.1"
-tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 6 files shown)


lib/src/core/fuzz/Cargo.toml
```diff
@@ -11,20 +11,12 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4"
 bytes = "1.0.1"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1.28", features = ["full"] }
```
0x2ec, line 14: not sure if specifying a minor version of tokio is desired - I can remove this.

locka99: Some users had problems with minor versions colliding with their own lock files so I took them off to make it easier for them. I don't think it matters on the fuzz file though

PR #288, opened 2023-09-07
Update crates version

cppcoffee:
Update dependent crates version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,7 +105,7 @@ dependencies = [
  "tokio-reactor",
  "tokio-tcp",
  "tokio-timer",
- "url",
+ "url 1.7.2",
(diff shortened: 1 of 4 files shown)


AiyionPrime (changes requested):
Hey @cppcoffee thanks for the effort,
though looking at the time and commits that have passed since "23, I think this needs a rebase or possibly a rework as whole.

Out of curiosity, have you documented or published your efforts of running OPC UA or rust in general on OpenWrt somewhere?

AiyionPrime:
@cppcoffee Are you still interested in this PR? Or can we close it?

cppcoffee:
It makes sense to submit a PR to fix the `gethostname` crate to support parsing OpenWrt hostnames.

PR #289, opened 2023-09-15
replaces tempdir with tempfile

antiblue:
Since the tempdir crate is being merged into [tempfile](https://github.com/Stebalien/tempfile) and is available in 3.x

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "opcua"
-version = "0.12.0" # OPCUARustVersion
+version = "0.12.0"                               # OPCUARustVersion
 description = "OPC UA client and server API"
 authors = ["Adam Lock <locka99@gmail.com>"]
 homepage = "https://github.com/locka99/opcua"
 repository = "https://github.com/locka99/opcua"
 license = "MPL-2.0"
-keywords = ["opcua","opc","ua"]
-categories = ["embedded","network-programming"]
+keywords = ["opcua", "opc", "ua"]
+categories = ["embedded", "network-programming"]
 readme = "../README.md"
 documentation = "https://docs.rs/opcua/"
 edition = "2021"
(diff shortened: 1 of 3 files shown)


lib/Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
 name = "opcua"
-version = "0.12.0" # OPCUARustVersion
+version = "0.12.0"                               # OPCUARustVersion
```
AiyionPrime, line 3: This looks unintended?

lib/Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
 description = "OPC UA client and server API"
 authors = ["Adam Lock <locka99@gmail.com>"]
 homepage = "https://github.com/locka99/opcua"
 repository = "https://github.com/locka99/opcua"
 license = "MPL-2.0"
-keywords = ["opcua","opc","ua"]
-categories = ["embedded","network-programming"]
+keywords = ["opcua", "opc", "ua"]
+categories = ["embedded", "network-programming"]
```
AiyionPrime, line 10: Maybe we could leave this change out of this PR and instead aim for something like toml-formatting?

lib/Cargo.toml
```diff
@@ -81,5 +81,5 @@ version = "0.7"
 optional = true
 
 [dev-dependencies]
-tempdir = "0.3"
+tempfile = "3.8"
```
AiyionPrime, line 84:
```suggestion
tempfile = "3.10"
```

AiyionPrime (changes requested):
While the PR looks good in general;
I think this PR lacks the respective change in the Cargo.lock?
And I'd avoid unrelated changes, especially if they are opinionated, like the amount of white spaces in line three.

lib/Cargo.toml
antiblue: yes, it was just the "autoformat on save"

lib/Cargo.toml
antiblue: yes, sorry about that

locka99:
I've manually merged this in, so closing the pull

PR #301, opened 2023-11-22
Several updates & Refactoring

flosse:
I will make ~~minor updates here as well as refactor the code a little.~~ several experiments by refactoring the client.
I would be happy if ~~these changes~~ ideas are well received and could be adopted.


.vscode/launch.json
```diff
@@ -29,7 +29,7 @@
             "program": "${workspaceRoot}/target/debug/opcua-simple-server.exe",
             "args": [],
             "stopAtEntry": false,
-            "cwd": "${workspaceRoot}/samples/simple-server",
+            "cwd": "${workspaceRoot}/exsamples/simple-server",
```
markussilvan: same typo

.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
             "program": "${workspaceRoot}/target/debug/opcua-simple-client.exe",
             "args": [],
             "stopAtEntry": false,
-            "cwd": "${workspaceRoot}/samples/simple-client",
+            "cwd": "${workspaceRoot}/exsamples/simple-client",
```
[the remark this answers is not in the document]
flosse: you're right but I removed the files anyway :wink:

See also:
https://github.com/locka99/opcua/blob/24162b9003431be1d467a3b365f08b6d8dfaf184/.gitignore#L1-L5

.vscode/launch.json
flosse: same here :relaxed:

markussilvan:
I think you should split these, so they can choose which parts they want or don't want to merge.

flosse:
> I think you should split these, so they can choose which parts they want or don't want to merge.

I assume that the maintainer is very busy and will not react at all for the time being.
In the meantime, I'll just play around a bit.
And if there is a more serious change, we can always discuss how this can be done :smile:

flosse:
I think this PR is not relevant anymore for this project.

Issue #341, opened 2024-06-18
regenerate internals in pipeline


AiyionPrime:
@locka99 Hey there,

first of all thanks for all the effort you've put into this project over the last years.
It appears to save me quite some time not needing to write an FFI for a cpp OPC UA library and hope you don't mind me rampaging through this project instead :S

Refactoring small stuff helps me figuring out how the project fits together;
~one thing I'm currently having trouble wrapping my head around is how and when the five schema helpers~

**update**
*I found they were all formatted and added the five respective PRs with necessary minor fixes below.*

- [x] `gen_address_space` #348
- [x] `gen_node_ids` #351
- [x] `gen_status_codes` #344
- [x] `gen_supported_message` #347 and
- [x] `gen_types` #350

~are invoked to match the current code committed.~

I suppose it should be a clean diff, when I regenerate and format the output?
If so I'd suggest (and eventually implement) a GitHub workflow, which regenerates the files and makes sure they do not change in order to tie the generators to the rest of the codebase?

This might furthermore help preventing unwanted fixes in generated code, which I might very well tripped in my open PRs.

I understand this is your freetime project? The open PRs are not meant as nudging, they only come in a bunch as I'll have more time to work on this in this week than usually. Sorry for all the buzz.

AiyionPrime:
After reading about it, I implemented the workflow for `gen_status_codes` in #344.

PR #351, opened 2024-06-27
ci: Regenerate and verify the node ids file

AiyionPrime:
> in order to enforce congruence between it and the source file.
> 
> Furthermore drop the generated file from rustfmt ignore as it's apparently currently formatted.


AiyionPrime:
As the last of the five generator PRs, this resolves #341.

PR #355, opened 2024-07-01
feat!: Provide cargo suited examples

### LINKED ISSUE #354: Example might better reside in an `examples` folder
Cargo has support for [examples](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) and using the conventional cargo way would reduce development time by about 15 seconds upon each `cargo build` on my machine.

Examples are automatically compiled upon tests, but not upon sole development.

AiyionPrime:
~Currently blocked by~

- [x] #368 or something that successfully updates the sample.

This PR drafts how to migrate the examples from `samples` workspaces to `lib/examples` to use cargos features for working and testing with them.

This eventually closes #354


AiyionPrime:
Currently the web server example is missing as it needs the update of actix-web, which @antiblue drafted a few months ago.

AiyionPrime:
That should be it.

AiyionPrime:
I expect each of these commits to pass the pipeline on its own.
@locka99 Should I split this into different PRs to migrate each sample on its own?

AiyionPrime:

PR #357, opened 2024-07-02
test: Implement an integration test for simple-server

### LINKED ISSUE #152: Simple-server failed to parse config
When running the `simple-server` sample code, I get the following error:

```
2022-01-09 18:08:18.062 - ERROR - opcua_core::config - Cannot deserialize configuration from ../server.conf
```

By inspecting the source code, I noticed the following line

[https://github.com/locka99/opcua/blob/0.8.1/core/src/config.rs#L42](https://github.com/locka99/opcua/blob/0.8.1/core/src/config.rs#L42)

This line reads the config from a YAML file and not from a config file as the one provided in the samples.

rustup 1.24.3 (ce5817a94 2021-05-31)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.57.0 (f1edd0429 2021-11-29)

Library version: 0.8.1

AiyionPrime:
> which verifies it passes configuration parsing and binds to the configured port in time.

This intends verify on a regular basis that the server fires up properly.
Which in turn closes #152, I think.

@lsabi I hope this test helps verifying your problem and prevent regressions,
but at the moment this appears to pass locally.

This test fires up the server, lets it bind the port, kills it and marks the test result accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1137,6 +1137,7 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "serde_yaml",
+ "socket2 0.5.6",
  "tempdir",
  "tokio",
  "tokio-util",
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -78,6 +78,7 @@ arc-swap = "1.6.0"
 [dev-dependencies]
 tempdir = "0.3"
 serde_json = "1.0"
+socket2 = "0.5.6"
 
 # Include console-logging when building tests
 opcua = { path = ".", features = ["console-logging"] }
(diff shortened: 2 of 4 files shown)


AiyionPrime:
@einarmo I hope to extend upon this tomorrow and spawn a client simultaneously and make sure its stdout contains value changes.

AiyionPrime:
https://github.com/locka99/opcua/actions/runs/9764628477/job/26953404110?pr=357#step:4:1415

AiyionPrime:
Looks like I need to mark OpenSSL in the Windows run, but Linux appears to be fine.
Any suggestions @locka99 ?

AiyionPrime:
The Cargo.lock appears to be unclean; I'll provide a CI step to prevent this.

AiyionPrime:
Might have been an issue during rebases; nevertheless #372 contains a check for `cargo build` induced changes.

AiyionPrime:

PR #360, opened 2024-07-17
feat: Upgrade `actix` and `actix-web`

### LINKED ISSUE #291: Upgrading actix-web
Hi all, I started working on upgrading the Actix-Web parts, but I stumble over an issue with Tokio. 

Through [`OPCUASession.connect`](https://github.com/locka99/opcua/blob/43e82f3f35ff86b50a84b6c2b5fdd0f112e21759/samples/web-client/src/main.rs#L203) the function [`get_server_endpoints_from_url<T>`](https://github.com/locka99/opcua/blob/8902bdb9b533d516f34ff935321deb8597a09f7f/lib/src/client/client.rs#L460) is called and at the end an instance of `Session` is dropped. 
This causes a panic, because Tokio wants to [enter a blocking region](https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/blocking/shutdown.rs#L44):

> thread 'actix-rt|system:0|arbiter:3' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', ...tokio-1.32.0\src\runtime\blocking\shutdown.rs:51:21

I have found [running-actix-web-using-tokiomain](https://docs.rs/actix-web/latest/actix_web/rt/index.html#running-actix-web-using-tokiomain) in Actix-Web's documentation, stating that `block_in_place` will not work. Unfortunately the backtrace reveals exactly that:

```
  13: core::ptr::drop_in_place<opcua::client::session::session::Session>
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be\library\core\src\ptr\mod.rs:497
  14: opcua::client::client::Client::get_server_endpoints_from_url<ref$<str$> >
             at ...\opcua\lib\src\client\client.rs:493
```

Any suggestions on how to fix this? 

(The code has been forked [here](https://github.com/LSAG-SID/opcua/tree/actix-web-upgrade))

AiyionPrime:
> from 0.7 and 0.7 to 0.13 and 4.8 respectively.
> 
> Furthermore add `actix-files` and `actix-web-actors` as new dependencies for functionalities which were extracted from `actix-web` in the past.

I think this eventually resolves #291.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,32 +4,27 @@ version = 3
 
 [[package]]
 name = "actix"
-version = "0.7.9"
+version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c616db5fa4b0c40702fb75201c2af7f8aa8f3a2e2c1dda3b0655772aa949666"
+checksum = "de7fa236829ba0841304542f7614c42b80fca007455315c45c785ccfa873a85b"
 dependencies = [
+ "actix-macros",
+ "actix-rt",
  "actix_derive",
- "bitflags 1.3.2",
- "bytes 0.4.12",
+ "bitflags 2.5.0",
+ "bytes",
  "crossbeam-channel",
- "failure",
- "fnv",
- "futures 0.1.31",
- "libc",
+ "futures-core",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
  "log 0.4.21",
- "parking_lot 0.7.1",
- "smallvec 0.6.14",
- "tokio 0.1.22",
- "tokio-codec",
- "tokio-executor",
- "tokio-io",
- "tokio-reactor",
- "tokio-signal",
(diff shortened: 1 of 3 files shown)


AiyionPrime:
As @vlnzrv pointed out in #291, this example will (without the upgrade) stop working with [rust 1.80.0](https://releases.rs/docs/1.80.0/) which is going to be released not in a couple of weeks, but **eight days**, if I'm not missing something else [here](https://github.com/rust-lang/rust/issues/125319#issuecomment-2203697058).

@locka99  I'll look into it later this afternoon, but it might be worthwhile to take half an hour in possibly the next week to review this, after I found a fix for the ungraceful runtime termination and tested this.

AiyionPrime:
While this update is conceptually a good idea, I found a less invasive way to ensure operability past Thursday.
I'll reopen this once I've got more time to spare.

PR #361, opened 2024-07-17
fix: Avert breakage with rust 1.80.0

AiyionPrime:
> By setting time 0.3.36 as minimally required version in accordance to https://github.com/rust-lang/rust/issues/125319#issuecomment-2120809372

@vlnzrv I think this is the least invasive resolution to the ongoing crate breakage due to the regression in 1.80.0.

Starting in eight days the opcua crate won't build in an up to date rust/stable anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,9 +11,16 @@ env:
 
 jobs:
   build-linux:
+    strategy:
+      matrix:
+        toolchain:
+          - stable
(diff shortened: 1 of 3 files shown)


AiyionPrime:
@locka99 This should reduce your time necessary to review the fix to a minimum.

AiyionPrime:
This has already been taken care of in a52f8f7.

PR #371, opened 2024-07-26
feat: Drop expendable drop

### LINKED ISSUE #370: `drop` without `Drop`
Hey @einarmo? What is the drop statement in line 73 supposed to do there?
Doesn't that only work on types implementing `Drop`?

The type of `activate_fut` is
`std::pin::Pin<&mut impl futures::Future<Output = std::result::Result<client::session::connect::SessionConnectMode, types::status_codes::StatusCode>>>`,

which is not `Drop`; does this do something I just don't get?

https://github.com/locka99/opcua/blob/8cc43e51fe2d14a965d271059aa031c89440fc30/lib/src/client/session/connect.rs#L44-L73

AiyionPrime:
> as it only extends lifetimes at this point and is likely a refactoring artifact.

If this does not break everything, this closes #370.
Thanks @einarmo for the fast explanation in the issue!

### DIFF
--- a/lib/src/client/session/connect.rs
+++ b/lib/src/client/session/connect.rs
@@ -70,8 +70,6 @@ impl SessionConnector {
             }
         };
 
-        drop(activate_fut);
-
         Ok((event_loop, id))
     }
 


AiyionPrime:
@einarmo Unless Windows starts breaking, this looks good to me?

PR #369, opened 2024-07-25
docs: Fix dead links

AiyionPrime:
@kwinz This intends to resolve #295.

@locka99 Everything but the broken web-client sample is documented properly, as its old actix version features structs not present in the more recent one documented for the main crate.
`cargo doc` does not handle different major versions with breaking changes clever enough.

This PR consists of atomic conventional commits (which could be cleanly rebased onto master), so the workaround (40ce6b6a75b6c79fb16be75a4f9e216cda2f8e92) can be reverted easily once #291 has been resolved.

PR #372, opened 2024-07-31
ci: Verify building does not induce repo changes

AiyionPrime:
> For example upon a change in Cargo.toml unreflected in Cargo.lock.


AiyionPrime:

Issue #205, opened 2022-05-13
The secure channel lifetime is hardcoded to 1 minute


jonasar:
The requested secure channel lifetime is hardcoded as 60 seconds in function issue_or_renew_secure_channel() in session_state.rs.

This is a request to add this as a parameter to the client builder instead.

I am using version 0.8.1 of the opcua library, but I see that this is still hardcoded in the latest commit on master.

jonasar:
This short hardcoded secure channel lifetime makes https://github.com/locka99/opcua/issues/207 much worse since the connection is lost approximately every 3rd time (in my case, connecting to the OPC UA server in a Siemens Siematic S7-1500) that the secure channel token is renewed.

BogdanYarotsky:
@AiyionPrime hi, I'm looking for a first issue to contribute. Do you think it's a good idea if I try to tackle this one? Seems not that complicated on a first glance.

AiyionPrime:
Good afternoon @BogdanYarotsky, first things first, the crates author and maintainer is Mr. Lock.
He should be a way better addressee for your request, as I'm currently just a passer-by either.

That said, I haven't looked into this nor #207 yet.
Skimming through both issues I think resolving #207 would be the one more worthwhile, as this one appears to just make living with #207 more unpleasant.

I'm usually a fan of fixing the root of problems and not relieving it of its symptoms.

I'll have to look into the spec, whether configurable channel lifetimes would serve a (in my eyes/subjectively) more worthwhile purpose than relief of #207 to answer this more properly.

I'll try coming back to this within the next days unless @locka99 finds the time for an assessment.

BogdanYarotsky:
Good evening @AiyionPrime, thank you for the quick reply! I've checked the spec and it seems that making the secure channel lifetime configurable can be beneficial on it it's own. Here is a description of this lifetime from the spec (OPC UA Specification Part 4 - Services):

`
The requested lifetime, in milliseconds, for the new SecurityToken. It
specifies when the Client expects to renew the SecureChannel by calling
the OpenSecureChannel Service again. If a SecureChannel is not
renewed, then all Messages sent using the current SecurityTokens shall
be rejected by the receiver.

Several cryptanalytic attacks become easier as more material encrypted
with a specific key is available. By limiting the amount of data processed
using a particular key, those attacks are made more difficult. Therefore
the volume of data exchanged between Client and Server shall be
limited by establishing a new SecurityToken after the lifetime.

[...]

PR #374, opened 2024-08-14
Configurable client's secure channel Lifetime

BogdanYarotsky:
Hello,

Here is a small PR based on the request in issue #205.

Now instead of hardcoded lifetime, the default value is passed from the client config (default is still 1 minute as it were with the previous const value so nothing is changed for existing users).

It's also now configurable using the ClientBuilder. Accepted value there is u32 and not a Duration to avoid implicit rounding of milliseconds.

I have added a few tests for a config builder but don't know how to test that the actual secure channel request is sending the right data. I would be grateful for suggestions.

In case of any questions please let me know.

Bogdan

### DIFF
--- a/lib/src/client/builder.rs
+++ b/lib/src/client/builder.rs
@@ -5,7 +5,6 @@ use crate::server::prelude::Config;
 
 use super::{Client, ClientConfig, ClientEndpoint, ClientUserToken, ANONYMOUS_USER_TOKEN_ID};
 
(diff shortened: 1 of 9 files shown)


einarmo (commented):
Testing the channel is challenging for sure, I have some thoughts of a way to expose some lower level details of the client in the future but for now there is no easy and unintrusive way to test this.

I think this is good, thanks.

PR #381, opened 2024-09-13
removed commented out code in symmetric_decrypt_verify

BogdanYarotsky:


AiyionPrime (changes requested):
@BogdanYarotsky Thanks for cleaning up :)

Any chance you could create a unittest from this commented out test code instead of dropping it and reference it in line 1213?

BogdanYarotsky:
@AiyionPrime Thanks for checking out the PR :)

I've brought back and uncommented the padding check. Also added a unit test to validate that works.

PR #396, opened 2024-10-29
Fixed connect not backing off when a connection attempt failed

guy176251:


lib/Cargo.toml
```diff
@@ -54,7 +54,7 @@ regex = "1.7"
 serde = "1.0"
 serde_derive = "1.0"
 serde_yaml = "0.9"
-serde_json = { version = "1.0", features = ["arbitrary_precision"] }
```
svanharmelen, line 57: Why would you remove this feature, what bug does it solve? This means you will likely loose precision which is what you want (at least I don't 😉).

guy176251: Was encountering a bug similar to the one in this issue:

https://github.com/serde-rs/serde/issues/2748

PR #399, opened 2024-11-05
Add some debug logging around the publish requests logic

svanharmelen:


svanharmelen:
@einarmo can I ask you about the reasoning of introducing `max_inflight_publish` versus the previous behavior of watching for `BadTooManyPublishRequests` messages and using that to limit the number of requests send?

Asking because I'm seeing issues due to the default value (2) and struggle to understand how to determine what a good value would be in our (or any actually) setup. So any background and/or pointers would be great. Thanks!

svanharmelen:
Closing this one as I want to propose a different solution.

PR #403, opened 2024-11-09
Make sure all timestamps remain consistent with one another

svanharmelen:
Please see the comment on #217 that explains in more detail why this change needs to be reverted in order to keep timestamps consistent.


einarmo (approved):
I agree that this is probably the correct way to do this.

PR #404, opened 2024-11-09
Clients should accept messages secured by an expired SecurityToken

svanharmelen:
I encountered a bug where an incoming message would fail with an invalid signature error. After quite some debugging I noticed this only occurred when the message was received just a few milliseconds after a security channel renewal.

After careful inspection it turned out that the message was still encrypted (which is according to spec) with the remote keys of the old security channel, but the client tried to decrypt and validate the message with the renewed keys.

According to the specs clients should accept messages secured by an expired SecurityToken for up to 25 % of the token lifetime (see https://reference.opcfoundation.org/Core/Part4/v105/docs/5.5.2 for more details), so I added a bit of logic to store the last 5 used remote keys and choose which one to use based on the token ID in the security header.

After making these changes and testing again against the same OPC UA server the issue did not occur again (while it occurred multiple times per hour before this fix was implemented) so this seems like a nice little improvement making the package a little bit more complaint.


lib/src/core/comms/secure_channel.rs
```diff
@@ -1048,8 +1071,18 @@ impl SecureChannel {
         self.local_keys.as_ref().unwrap()
     }
 
-    fn remote_keys(&self) -> &(Vec<u8>, AesKey, Vec<u8>) {
-        self.remote_keys.as_ref().unwrap()
+    fn insert_remote_keys(&mut self, keys: (Vec<u8>, AesKey, Vec<u8>)) {
+        self.remote_keys[self.token_id as usize % STORED_REMOTE_KEYS] = Some(keys);
```
einarmo: This approach, while convenient, appears to assume that `token_id`s are sequential. Is that really guaranteed?

lib/src/core/comms/secure_channel.rs
```diff
@@ -35,6 +35,9 @@ pub enum Role {
     Server,
 }
 
+// Number of stored remote keys.
+const STORED_REMOTE_KEYS: usize = 5;
```
einarmo: Is it really possible for more than two tokens to be live at the same time? From reading the standard it seems like having three or more tokens overlapping would be impossible.

einarmo (commented):
This is clearly something we need, but I'm not entirely sold on the approach, see comments.

lib/src/core/comms/secure_channel.rs
einarmo: It might be easier to just manually track `last_token_id` and `last_remote_key`, or something like that.

lib/src/core/comms/secure_channel.rs
svanharmelen: Yeah fair point, I guess that is not defined in the spec and so not guaranteed.

lib/src/core/comms/secure_channel.rs
svanharmelen: Not sure if that is possible but the spec mentions clients should accept messages secured by an expired SecurityToken for up to 25 % of the token lifetime. So not sure if it can happen that rapid renews (for whatever reason) could occur within that time.

I looked at how other packages solved this and they generally seem to use a map where the token ID is the key and the value are the keys.

Initially I started out with a map as well, but if you want to keep that from growing endlessly you also need to rely on the token ID being ordered. Which is again not part of the spec. It's only required for them to be unique.

But I do want to make it robust and agree that currently it isn't. Have an other idea, will have another look tomorrow...

einarmo: We are the ones that renew tokens, I believe, so we do actually have complete control over what overlap is allowed. Since we only ever renew once 75% of the token lifetime has expired, there can never be more than two active tokens at a time.

svanharmelen:
Updated the PR... It maybe just a little bit more involved, but it follows the spec for 100% this way.

Note that I removed the `token_has_expired` method as it contained a bug (used the lifetime as seconds instead of milliseconds) and wasn't used by any of the client or server code.

lib/src/core/comms/secure_channel.rs
```diff
@@ -1048,8 +1066,33 @@ impl SecureChannel {
...
+        let expires_at = (self.token_lifetime as f32 * 1.25).ceil();
+        let expires_at = Duration::milliseconds(expires_at as i64);
+
+        // Then insert the new keys to ensure there is
+        // always at least one set of keys available.
+        self.remote_keys.insert(
+            self.token_id,
+            RemoteKeys {
+                keys,
+                expires_at: self.token_created_at + expires_at,
```
einarmo: Without looking too closely at the code, it seems like `token_created_at` is subject to some internal invariants that could be better modeled (it must correspond to the expires_at in the remote_keys entry given by `token_id`). Perhaps a target for some future refactoring...

Not a blocker.

lib/src/core/comms/secure_channel.rs
```diff
@@ -35,6 +36,12 @@ pub enum Role {
     Server,
 }
 
+#[derive(Debug)]
+struct RemoteKeys {
+    keys: (Vec<u8>, AesKey, Vec<u8>),
+    expires_at: DateTime,
```
einarmo, line 42: Avoid using `DateTime` for timestamps, if possible, confusing as that may sound.

If time is purely relative, always prefer using `std::time::Instant` and `std::time::Duration`, these generally point to the monotonic high frequency clock, while clock time like `SystemTime` or those provided by `chrono` use the system clock, which can suddenly change for a number of reasons.

lib/src/core/comms/secure_channel.rs
```diff
@@ -1048,8 +1066,33 @@ impl SecureChannel {
...
 
-    fn remote_keys(&self) -> &(Vec<u8>, AesKey, Vec<u8>) {
-        self.remote_keys.as_ref().unwrap()
+    fn insert_remote_keys(&mut self, keys: (Vec<u8>, AesKey, Vec<u8>)) {
+        // First remove any expired keys.
+        self.remote_keys
+            .retain(|_, v| DateTime::now() < v.expires_at);
+
+        let expires_at = (self.token_lifetime as f32 * 1.25).ceil();
+        let expires_at = Duration::milliseconds(expires_at as i64);
```
einarmo: Nit: Name is confusing, `expires_in` is probably more accurate.

einarmo (commented):
I think this is a perfectly fine solution, but see my comment on timestamps. We should strongly prefer using std time for any internal mechanisms.

lib/src/core/comms/secure_channel.rs
svanharmelen: Fair point, will change it to `Instant`

svanharmelen: Hmm... After taking a closer look I don't think that makes sense here. The expiration time is calculated based on the time the token was created. That creation time comes from the server in the form of a timestamp and is most likely in the past by the time we get here.

So somehow converting that to an `Instant` doesn't make much sense to me especially taking into account that there are lots and lots of other places in the code that do use a `DateTime`. Which in most cases makes sence because these values are sometimes also used in messages between the client and the server.

Of course we could so something like the pseudo code pasted below, but I honestly don't see much benefit from that:

```rust
let expires_in = Instant::now() + (DateTime::now() - (self.token_created_at + expires_in));
```

While the argument about the system clock if very true, in the case of this crate a whole lot more will break (for example the secure channel) if the system clock got updated while the program is running. So unless you are thinking about another solution or have more insights to share, I think I prefer to keep it like this.

lib/src/core/comms/secure_channel.rs
svanharmelen: I'm not sure I understand what you mean with `token_created_at is subject to some internal invariants`? Both the token ID and associated created at time are set using the same call during the issue or renew of a channel.

lib/src/core/comms/secure_channel.rs
einarmo: I didn't recall that it was received from the server, doing it that way is _usually_ suboptimal, but OPC-UA isn't always optimal... If so then you're right, we need to use DateTime, with all its potential issues.

lib/src/core/comms/secure_channel.rs
einarmo: As I said, not a big deal, my point is just that there's now a built-in relationship in the `SecureChannel` type between `token_id`, `token_created_at`, and `RemoteKeys.expires_at`, it would be better to not have the `token_created_at` parameter at all, and instead store it only inside the `remote_keys` map.

einarmo: The whole `SecureChannel` type has a few of these invariants that could be modeled in the type system, IMO. It can be done in a future PR, I'm fine with this as is now.

svanharmelen: Check, got it 👍🏻

PR #417, opened 2024-11-30
Add CancellationToken as argument to server methods

oroulet:
This is the advice tokio way to stop a task


oroulet:
This is not really finished but works for me. I am open to any suggestion to improve API

PR #425, opened 2025-04-01
Adapt the crate for puffin (make it Mapper)

LCBH:
Goal is to minimize modifications, using annotations as much as possible to ease future merges from upstream.


LCBH:
Sorry, I created the PR in the wrong place.

